Issue #35, opened 2017-01-20
Responsiveness of the platform - pt. 2


canihavesomecoffee:
There are some smaller issues with the responsiveness of the platform:

- On mobile, the dropdown for the categories is too wide. Better styling is needed
- ~The table of latest tests isn't any good on mobile (maybe we should use a responsive table plugin?)~ (fixed in #45)
- ~Input type of login email field should be email instead of text~ (fixed in #43)

canihavesomecoffee:
Worth 1 point for GSoC Quali.

satyammittal:
i would like to work on issue

canihavesomecoffee:
Just do so, and make a PR :)

PR #43, opened 2017-02-10
Email form type changed to email

satyammittal:
#35 
fixed-"Input type of login email field should be email instead of text"

### DIFF
--- a/mod_auth/forms.py
+++ b/mod_auth/forms.py
@@ -10,14 +10,14 @@ def unique_username(form, field):
     # Check if a user already exists with this name
     user = User.query.filter(User.name == field.data).first()
     if user is not None:
-        raise ValidationError('there is already a user with this name')
+        raise ValidationError('There is already a user with this name')
 
 
 def valid_password(form, field):
     if len(field.data) == 0:
         raise ValidationError('new password cannot be empty')
     if len(field.data) < 10 or len(field.data) > 500:
-        raise ValidationError('password needs to be between 10 and 500 '
+        raise ValidationError('Password needs to be between 10 and 500 '
                               'characters long (you entered %s characters'
                               % len(field.data))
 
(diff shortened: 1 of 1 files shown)


mod_auth/forms.py
```diff
@@ -42,13 +42,13 @@ class LoginForm(Form):
     email = StringField('Email', [DataRequired(
         message='Email is not filled in.')])
     password = PasswordField('Password', [
-        DataRequired(message='Password cannot be empty.')])
+        DataRequired(message='Password cannot be empty.'), Email("entered value is not a valid email address")])
```
saurabhshri: You probably meant to add it on #L43.
Also, please change it from `entered` to `Entered` while you are at it.

mod_auth/forms.py
```diff
@@ -42,13 +42,13 @@ class LoginForm(Form):
...
     submit = SubmitField('Login')
 
 
 class SignupForm(Form):
     email = StringField('Email', [DataRequired(
-        message='Email is not filled in.')])
+        message='Email is not filled in.'), Email("entered value is not a valid email address")])
```
saurabhshri: Same change as above.

mod_auth/forms.py
satyammittal: oh yes !!
Thanks

canihavesomecoffee:
Fixes are just adding extra validation (which isn't bad on itself), but do not change the type to email, as requested in the issue.

satyammittal:
then should i only have to validate for login?

saurabhshri:
@satyammittal I think you misunderstood the requirement.  Currently, the input type in the generated  HTML form is of type _text_. You simply have to change it to _email_ wherever required. 


email


It has already been done in the same file you're working. Just look at it. It's just a matter of changing from `StringField` to `EmailField`.

PR #48, opened 2017-02-15
Change dropdown width for mobile

### LINKED ISSUE #35: Responsiveness of the platform - pt. 2

madan96:
Adds device specific styling for categories dropdown menu. 

Closes #35 .

### DIFF
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -337,3 +337,11 @@ summary {
 table.stack td {
     border-top:0;
     word-break:break-word; }
+
+@media only screen and (max-device-width: 480px) {
+  .off-canvas.position-left {
+    left: -250px;
+    top: 0;
+    width: 200px;
+    }
+}


canihavesomecoffee:
I don't really see any styling for dropdown's at all... Maybe you can explain what the change does?

madan96:
For now, I just adjusted the width of the menu in case of mobiles. I'll try adding a Sidebar overlay feature for mobiles.

canihavesomecoffee:
Clarification: I just want the arrow for the regression test category dropdown (on the page, nothing to do with the menu) to be visible at all times (not partially out of view). Please make a new pull request with that fix.

PR #50, opened 2017-02-18
Dropdown css changed

satyammittal:
fixed [#35]
screen shot 2017-02-18 at 18 12 55
screenshot from 2017-02-18 18-13-04


canihavesomecoffee:
There's no need for a restyle, or having to use a boat-load of javascript. If the arrow is displayed nicely instead of half-cropped on certain view sizes it's fine already.

satyammittal:
ok
i will do that

PR #56, opened 2017-03-01
added documentation for mod_auth/models.py

pranavsid98:
Documentation for mod_auth/models.py

### DIFF
--- a/mod_auth/models.py
+++ b/mod_auth/models.py
@@ -23,8 +23,9 @@ class User(Base):
     role = Column(Role.db_type())
 
     def __init__(self, name, role=Role.user, email=None, password=''):
-	'''
+	"""
 	Parametrized constructor for the User model
+
 	:param name: The value of the 'name' field of User model
 	:type name: str
 	:param role: The value of the 'role' field of User model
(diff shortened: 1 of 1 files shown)


mod_auth/models.py
```diff
@@ -23,35 +23,70 @@ class User(Base):
     role = Column(Role.db_type())
 
     def __init__(self, name, role=Role.user, email=None, password=''):
+	'''
```
canihavesomecoffee: Please use double quotes instead of single ones (ref. PEP257)

mod_auth/models.py
```diff
@@ -23,35 +23,70 @@ class User(Base):
...
+	Parametrized constructor for the User model
```
canihavesomecoffee: I'm missing the arguments of the constructor.

mod_auth/models.py
```diff
@@ -23,35 +23,70 @@ class User(Base):
...
+	Represent a User Model by its 'name' Field.
+	'''
         return '<User %r>' % self.name
 
     @staticmethod
     def generate_hash(password):
+	'''
+	Generates a Hash value for a password
+	:param password: The password to be hashed
+	:type password: str
```
canihavesomecoffee: :return and :rtype are missing. They provide some help when using an IDE. Please add for all methods that return something.

pranavsid98:
Requested changes made.

pranavsid98:
Requested changes made

mod_auth/models.py
```diff
@@ -23,35 +23,98 @@ class User(Base):
     role = Column(Role.db_type())
 
     def __init__(self, name, role=Role.user, email=None, password=''):
+	"""
+	Parametrized constructor for the User model
+
+	:param name: The value of the 'name' field of User model
+	:type name: str
+	:param role: The value of the 'role' field of User model
+	:type role: Role Model
```
canihavesomecoffee: Almost. Mention Role should be enough.

mod_auth/models.py
```diff
@@ -23,35 +23,98 @@ class User(Base):
...
+	:return : Compares role of the passed Model with 'Admin' and returns result	
+	:rtype: boolean
+	"""
         return self.role == Role.admin
 
     def has_role(self, name):
+	"""
+	Checks whether the User has a particular role
+
+	:param name: Name of the user
```
canihavesomecoffee: Role to check, not he name of the user

mod_auth/models.py
```diff
@@ -23,35 +23,98 @@ class User(Base):
...
+	:type name: str
+	:return : Checks whether a passed User has a role.
```
canihavesomecoffee: Not fully ;) We have a user object, and call it like user.has_role('admin'), so this phrase isn't really correct.

mod_auth/models.py
```diff
@@ -23,35 +23,98 @@ class User(Base):
...
+	:type new_password: str
+	"""
         self.password = self.generate_hash(new_password)
 
     @property
     def is_admin(self):
+	"""
+	Verifies if a User is the admin
+
+	:return : Compares role of the passed Model with 'Admin' and returns result
```
canihavesomecoffee: We don't pass anything. We just check if this user has an admin role or not.

pranavsid98:
Requested changes made

PR #76, opened 2017-06-18
[FIX] added windows dual support

satyammittal:

- [x] I have read and understood the [contributors guide](https://github.com/canihavesomecoffee/sample-platform/blob/master/.github/CONTRIBUTING.md).

---

added support for both testing in both linux and windows in parallel

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -153,7 +153,7 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
         full_url = url_for('ci.progress_reporter', test_id=test.id,
                            token=test.token, _external=True, _scheme="https")
     file_path = os.path.join(config.get('SAMPLE_REPOSITORY', ''),
-                             kvm_name , 'reportURL')
+                             kvm_name, 'reportURL')
     with open(file_path, 'w') as f:
         f.write(full_url)
     # 1) Generate test files
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -230,7 +230,8 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
     # 2) Create git repo clone and merge PR into it (if necessary)
     try:
         repo = Repo(os.path.join(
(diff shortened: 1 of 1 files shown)


mod_ci/controllers.py
```diff
@@ -152,12 +152,13 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
     with app.app_context():
         full_url = url_for('ci.progress_reporter', test_id=test.id,
                            token=test.token, _external=True, _scheme="https")
-    file_path = os.path.join(config.get('SAMPLE_REPOSITORY', ''), 'reportURL')
+    file_path = os.path.join(config.get('SAMPLE_REPOSITORY', ''),
+                             kvm_name + '/reportURL')
```
canihavesomecoffee: please just add the kvm name as an argument to the join... join accepts multiple paths to be joined together.

mod_ci/controllers.py
```diff
@@ -152,12 +152,13 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
...
     with open(file_path, 'w') as f:
         f.write(full_url)
     # 1) Generate test files
     base_folder = os.path.join(
-        config.get('SAMPLE_REPOSITORY', ''), 'ci-tests')
+        config.get('SAMPLE_REPOSITORY', ''), kvm_name + '/ci-tests')
```
canihavesomecoffee: Same remark

mod_ci/controllers.py
```diff
@@ -313,7 +315,7 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
         #     return
         # TODO: check what happens on merge conflicts
     else:
-        test_branch = repo.create_head('CI_Branch', 'HEAD')
+        test_branch = repo.create_head('CI_Branch_' + kvm_name, 'HEAD')
```
canihavesomecoffee: This method won't work. We don't need separate branches (the VM's don't care about git), but separate directories with separate git clones.

PR #106, opened 2017-08-31
[FIX] fixed some issues in sample upload regarding revoke

satyammittal:

- [x] I have read and understood the [contributors guide](https://github.com/canihavesomecoffee/sample-platform/blob/master/.github/CONTRIBUTING.md).

---

{pull request content here}
checking the validity of token and updating it if required.

### DIFF
--- a/mod_auth/controllers.py
+++ b/mod_auth/controllers.py
@@ -123,9 +123,8 @@ def github_token_validity(token):
     :param token: The value of 'github_token' stored in the user
         model
     :type token: str
-    return bool: Returns True/False based on the validity of the
-        token.
-    rtype: bool
+    :return True/False: Returns whether token is valid or not
+    :rtype: bool
     """
     from run import config
     github_clientid = config.get('GITHUB_CLIENT_ID', '')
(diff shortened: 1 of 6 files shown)


mod_auth/controllers.py
```diff
@@ -117,6 +117,26 @@ def send_reset_email(usr):
               'error-message')
 
 
+@mod_auth.route('/github_token_validity', methods=['GET', 'POST'])
```
canihavesomecoffee: If this is an internal method, there's no need to have a route assigned to it.

mod_auth/controllers.py
```diff
@@ -117,6 +117,26 @@ def send_reset_email(usr):
...
+def github_token_validity(token):
+    """
+    Check token validity
+    Set token to null if not valid
```
canihavesomecoffee: This def does not change the token of a user.

Argument type & description are missing, just as the return values.

mod_upload/controllers.py
```diff
@@ -325,6 +325,10 @@ def process_id(upload_id):
                             )
                             g.db.add(issue)
                             g.db.commit()
+                        else:
+                            flash('Could not submit an issue on GitHub '
```
canihavesomecoffee: You add a flash message, but I see no changes to the templates to show the flashed message.

Issue #123, opened 2017-12-13
[BUG] Github status not updated when tests completed


satyammittal:
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

Sample platform commit (found at the bottom of each page) : **X.X.X**

**In raising this issue, I confirm the following (please check boxes, eg [X]):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/sample-platform/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/sample-platform/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/sample-platform/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.

**My familiarity with the project is as follows (check one, eg [X]):**
- [ ] I have never visited/used the platform.
- [ ] I have used the platform just a couple of times.
- [ ] I have used the platform extensively, but have not contributed previously.
- [x] I am an active contributor to the platform.

---
{issue content here, replace this line with your issue content}
On pull request 835
![image](https://user-images.githubusercontent.com/15922592/33923281-45a4c40a-dff6-11e7-8271-dbcd8c572529.png)
On pull request 822
![image](https://user-images.githubusercontent.com/15922592/33923320-882f88e6-dff6-11e7-9f86-e7a2fea0c78d.png)
We can see that on both completion of linux and windows testing,
Github test status is not getting updated every-time.

canihavesomecoffee:
Seems to be fixed 🥇

PR #125, opened 2017-12-13
Rework some code to potentially fix the non-posting issue

canihavesomecoffee:
- Removes the commit from the PR's if it's no longer applicable
- No longer keeps track of the GH status if the commit is empty
- Trigger new tests only after posting current status instead of otherwise.


canihavesomecoffee:
Potentially fixes #123

PR #144, opened 2018-01-02
[FEATURE] Stop Blocked users from queing their Pull Request

MatejMecka:

---


mod_ci/controllers.py
```diff
@@ -10,7 +10,7 @@
 import os
 import shutil
 import sys
-
+import requets
```
TheClashster: This should be requests btw 😃

satyammittal: @MatejMecka also fix the PEP8 issues

mod_ci/controllers.py
```diff
@@ -62,7 +62,8 @@ def before_app_request():
         ]
     )
     if 'config' in g.menu_entries and 'entries' in config_entries:
-        g.menu_entries['config']['entries'] = config_entries['entries'] + g.menu_entries['config']['entries']
+        g.menu_entries['config']['entries'] = config_entries['entries'] + \
```
canihavesomecoffee: Why again all these changes? Please stop changing these **unneeded** things

mod_ci/controllers.py
```diff
@@ -463,33 +463,48 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+    fork_id = Test.query.filter(Test.fork_id == Fork.id).first()
+    githubapi = requests.get(
+        "https://api.github.com/repos/ccextractor/ccextractor/pulls/{}".format())
+    userjson = json.loads(githubapi.body)
+    user = userjson['user']['id']
+
+    record_from_db = BlockedUsers.query.filter_by(userID=user).first()
+
+    if user == record_from_db.userID:
+        log.critical("Error. User in Blacklist!")
```
canihavesomecoffee: So besides a log entry (who no-one can see besides me) there's no trace whatsoever that something happened at all. The user doesn't get notified either, which is required.

mod_ci/controllers.py
MatejMecka: I don't know how I'm doing that. I swear to god I'm doing this by accident and not noticing them

mod_ci/controllers.py
```diff
@@ -445,7 +446,6 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
 def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     """
     Function to store test details into Test model for each platform, and post the status to GitHub.
-
```
canihavesomecoffee: Please re-add this line.

mod_ci/controllers.py
```diff
@@ -463,23 +463,45 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     """
     from run import log
 
-    fork = Fork.query.filter(Fork.github.like("%/CCExtractor/ccextractor.git")).first()
+    fork = Fork.query.filter(Fork.github.like(
```
canihavesomecoffee: Please undo.

mod_ci/controllers.py
```diff
@@ -463,23 +463,45 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+        "%/CCExtractor/ccextractor.git")).first()
+    fork_id = Test.query.filter(Test.fork_id == Fork.id).first()
```
canihavesomecoffee: Did you test this? As far as I can see, this will return the first instance of a Test that satisfied a condition that won't work either, not a fork id.

mod_ci/controllers.py
```diff
@@ -463,23 +463,45 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+    githubapi = requests.get(
+        "https://api.github.com/repos/ccextractor/ccextractor/pulls/{}".format())
```
canihavesomecoffee: Empty format?

mod_ci/controllers.py
```diff
@@ -463,23 +463,45 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+    userjson = json.loads(githubapi.body)
+    user = userjson['user']['id']
+
+    record_from_db = BlockedUsers.query.filter_by(userID=user).first()
+    blocked = False
 
-    if test_type == TestType.pull_request:
-        branch = "pull_request"
+    if user == record_from_db.userID:
```
canihavesomecoffee: This won't work if there's no matching user, because first will return None, and you can't get a property from that.

mod_ci/controllers.py
```diff
@@ -463,23 +463,45 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+            }
+
+    if blocked is True:
+        gh_commit.post(
+            state=Status.ERROR,
+            description="Tests cancelled! You may be blocked.",
+            context="CI - {name}".format(name=platform_name),
+            target_url=url_for('test.by_id', test_id=test_id, _external=True)
+        )
+    else:
```
canihavesomecoffee: This if-else makes no sense. Why not return early when there's a match for a blocked user, so that the remaining flow can be left untouched?

MatejMecka:
ignore

mod_ci/controllers.py
```diff
@@ -464,22 +465,39 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     from run import log
 
     fork = Fork.query.filter(Fork.github.like("%/CCExtractor/ccextractor.git")).first()
+    test_case = Test.query.filter(Test.fork_id == Fork.id).first()
```
canihavesomecoffee: This still won't work. You're currently comparing a fork_id of the Test to any Fork id... That makes no sense. Didn't you intend to use the `fork` object instead of the `Fork` class?

canihavesomecoffee: But why all the effort to retrieve a test case (which wasn't added yet!) when the pr_nr (in case of a pull request), is passed to the method?

mod_ci/controllers.py
```diff
@@ -464,6 +465,22 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     from run import log
 
     fork = Fork.query.filter(Fork.github.like("%/CCExtractor/ccextractor.git")).first()
+    githubapi = requests.get("https://api.github.com/repos/ccextractor/ccextractor/pulls/{pr}".format(pr=Test(pr_nr)))
```
canihavesomecoffee: Why are you still using `Test(pr_nr)`? We pass the `pr_nr` in as an argument to this function...

mod_ci/controllers.py
```diff
@@ -464,6 +465,22 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+    userjson = json.loads(githubapi.body)
+    user = userjson['user']['id']
+
+    record_from_db = BlockedUsers.query.filter_by(userID=user).first()
+
+    if user == record_from_db.userID:
```
canihavesomecoffee: What if record_from_db is None? That will happen if there's no matching user...

mod_ci/controllers.py
```diff
@@ -464,6 +465,22 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+        log.critical("Error. User in Blacklist!")
+        gh_commit.post(
+            state=Status.ERROR,
+            description="Tests cancelled! You may be blocked.",
+            context="CI - {name}".format(name=platform_name),
```
canihavesomecoffee: Platform name is undefined.

mod_ci/controllers.py
```diff
@@ -464,6 +465,22 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+            target_url=url_for('test.by_id', test_id=test_id, _external=True)
```
canihavesomecoffee: test_id is undefined.

mod_ci/controllers.py
```diff
@@ -480,16 +497,17 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
-                gh_commit.post(
-                    state=Status.PENDING,
-                    description="Tests queued",
-                    context="CI - {name}".format(name=platform_name),
-                    target_url=url_for('test.by_id', test_id=test_id, _external=True)
-                )
-            except ApiError as a:
-                log.critical('Could not post to GitHub! Response: {res}'.format(res=a.response))
+
+    for platform_name, test_id in status_entries.items():
```
canihavesomecoffee: Why is the indentation changed?

mod_ci/controllers.py
```diff
@@ -464,6 +465,11 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     from run import log
 
     fork = Fork.query.filter(Fork.github.like("%/CCExtractor/ccextractor.git")).first()
+    githubapi = requests.get("https://api.github.com/repos/ccextractor/ccextractor/pulls/{pr}".format(pr=pr_nr))
```
canihavesomecoffee: What if pr_nr is 0 (a.k.a. when it's a commit instead of a PR)

mod_ci/controllers.py
```diff
@@ -474,6 +480,17 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     db.add(windows_test)
     db.commit()
 
+    if user == record_from_db.userID is not None:
```
canihavesomecoffee: This still won't work.

mod_ci/controllers.py
```diff
@@ -474,6 +480,17 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+        for platform_name, test_id in status_entries.items():
```
canihavesomecoffee: `status_entries` is not defined! Please check your code!

mod_ci/controllers.py
```diff
@@ -464,9 +465,18 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     from run import log
 
     fork = Fork.query.filter(Fork.github.like("%/CCExtractor/ccextractor.git")).first()
+    github_apiurl = "https://api.github.com/repos/ccextractor/ccextractor/pulls/{pr}"
+    userjson = ""
+    user = ""
 
     if test_type == TestType.pull_request:
         branch = "pull_request"
+        github_apiurl.format(pr=pr_nr)
```
canihavesomecoffee: Why split definition and formatting if it's only used here anyway?

mod_ci/controllers.py
```diff
@@ -464,9 +465,18 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
+        res_from_github = requests.get(github_apiurl)
+        userjson = json.loads(res_from_github.body)
```
canihavesomecoffee: Same with this.

mod_ci/controllers.py
```diff
@@ -481,6 +491,15 @@ def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
...
         for platform_name, test_id in status_entries.items():
+            if record_from_db is not None:
+                log.critical("Error. User in Blacklist!")
+                gh_commit.post(
+                   state=Status.ERROR,
+                   description="Tests cancelled! You may be blocked.",
+                   context="CI - {name}".format(name=platform_name),
+                   target_url=url_for('test.by_id', test_id=test_id, _external=True)
+                   )
+                return "Blocked User!"  # https://www.youtube.com/watch?v=AOHy4Ca9bkw
```
canihavesomecoffee: The tests are left behind, and will be executed anyway next time another one is queued. That's not good.

canihavesomecoffee:
Closing this as there's still no sight for a correct solution, and another one was merged in the meanwhile.

PR #145, opened 2018-01-02
[FEATURE] Implementing the BlockedUsers Data Model

TheClashster:

---

Have made it pull request specific, by comparing the ID of the user with the one stored in the data model

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -62,8 +62,7 @@ def before_app_request():
         ]
     )
     if 'config' in g.menu_entries and 'entries' in config_entries:
-        g.menu_entries['config']['entries'] = \
-            config_entries['entries'] + g.menu_entries['config']['entries']
+        g.menu_entries['config']['entries'] = config_entries['entries'] + g.menu_entries['config']['entries']
     else:
         g.menu_entries['config'] = config_entries
 
@@ -445,8 +444,7 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
 
 def queue_test(db, gh_commit, commit, test_type, branch="master", pr_nr=0):
     """
-    Function to store test details into Test model separately for each platform.
-    Post status to GitHub.
+    Function to store test details into Test model for each platform, and post the status to GitHub.
 
     :param db: Database connection.
     :type db: sqlalchemy.orm.scoped_session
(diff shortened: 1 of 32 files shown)


mod_ci/controllers.py
```diff
@@ -444,52 +445,39 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
...
+        "%/CCExtractor/ccextractor.git")).first()
     if test_type == TestType.pull_request:
         branch = "pull_request"
-
-    linux_test = Test(TestPlatform.linux, test_type, fork.id, branch, commit, pr_nr)
-    db.add(linux_test)
-    windows_test = Test(TestPlatform.windows, test_type, fork.id, branch, commit, pr_nr)
-    db.add(windows_test)
+    # Create Linux test entry
+    linux = Test(TestPlatform.linux, test_type, fork.id, branch, commit,
```
satyammittal: You don't have to short each line. You can kept line length upto 90 characters. It will be easy to read.

TheClashster: Are you sure? The PEP8 guidelines say 79 characters.

satyammittal: Yes but on sample platform we have increased it to 90 characters, you can simply try pushing the changes upto 90 characters line.

satyammittal: We already simplified all lines which are exceeding 79 characters in precious commit:
https://github.com/CCExtractor/sample-platform/commit/8403bae4583da989a96633ef6c05844ae6e5ecb1

TheClashster: Will send another commit to that accord😃

satyammittal:
Better be just undo all the changes of line length you have done in your commit. That is not useful.

mod_ci/controllers.py
```diff
@@ -555,43 +555,48 @@ def start_ci():
...
-                        description="Tests canceled",
-                        context="CI - {name}".format(name=test.platform.value),
-                        target_url=url_for('test.by_id', test_id=test.id, _external=True)
-                    )
-            elif payload['action'] == 'reopened':
-                # Run tests again
-                queue_test(g.db, gh_commit, commit, TestType.pull_request, pr_nr=pr_nr)
+            # Check if user blacklisted
+            if BlockedUsers.query.filter(BlockedUsers.userID == payload[
+                   'pull_request']['user']['id']).first():
```
canihavesomecoffee: No trace besides a single log line, and user doesn't get notified? Please change.

TheClashster: Added return.

mod_ci/controllers.py
```diff
@@ -555,43 +555,55 @@ def start_ci():
...
+                gh_commit.post(
+                    state=Status.ERROR,
+                    description="CI start aborted. You may be blocked from accesing this functionality",
+                    context="CI - {name}".format(name=platform_name),
+                    target_url=url_for('test.by_id', test_id=test_id, _external=True)
                     )
-            elif payload['action'] == 'reopened':
-                # Run tests again
-                queue_test(g.db, gh_commit, commit, TestType.pull_request, pr_nr=pr_nr)
+                return 'ERROR'
```
canihavesomecoffee: If you do a return here, you can drop the else and un-indent the entire other block. That'll also reduce the diff, and make it easier to maintain ;)

TheClashster: Will do 😃

mod_ci/controllers.py
```diff
@@ -555,6 +555,17 @@ def start_ci():
 
         elif event == "pull_request":
             # If it's a PR, run the tests
```
canihavesomecoffee: Maybe move this comment to after your changes, that'll make more sense.

mod_ci/controllers.py
```diff
@@ -555,6 +555,17 @@ def start_ci():
...
+            # Check if user blacklisted
+            if BlockedUsers.query.filter(BlockedUsers.userID == payload[
+                   'pull_request']['user']['id']).first():
+                log.critical("User Blacklisted")
+                gh_commit.post(
+                    state=Status.ERROR,
+                    description="CI start aborted. You may be blocked from accesing this functionality",
+                    context="CI - {name}".format(name=platform_name),
```
canihavesomecoffee: This name doesn't seem to be defined?

mod_ci/controllers.py
TheClashster: One second.

mod_ci/controllers.py
```diff
@@ -554,7 +554,17 @@ def start_ci():
...
         elif event == "pull_request":
-            # If it's a PR, run the tests
+            # Check if user blacklisted
+            if BlockedUsers.query.filter(BlockedUsers.userID == payload[
+                   'pull_request']['user']['id']).first():
+                log.critical("User Blacklisted")
+                gh_commit.post(
+                    state=Status.ERROR,
+                    description="CI start aborted. You may be blocked from accesing this functionality",
+                    target_url=url_for(@mod_ci.route('/'), _external=True)
```
canihavesomecoffee: No, that's not the correct way to link.

See https://github.com/CCExtractor/sample-platform/blob/8403bae4583da989a96633ef6c05844ae6e5ecb1/mod_auth/controllers.py#L53

satyammittal:
@TheClashster Update your branch with latest changes

mod_ci/controllers.py
```diff
@@ -564,6 +564,17 @@ def start_ci():
                 except KeyError:
                     g.log.critical("Didn't find a SHA value for a newly opened PR!")
                     g.log.debug(payload)
+                # Check if user blacklisted
+                if BlockedUsers.query.filter(BlockedUsers.userID == payload[
+                    'pull_request']['user']['id']).first() is not None:
+                    log.critical("User Blacklisted")
```
saurabhshri: I guess it'd be helpful to also log user ID here.

mod_ci/controllers.py
```diff
@@ -564,6 +564,17 @@ def start_ci():
...
+                    gh_commit.post(
+                        state=Status.ERROR,
+                        description="CI start aborted. You may be blocked from accesing this functionality",
```
saurabhshri: Typo - "accessing".

Also maybe change it to something short like "Tests aborted - user is blacklisted.". What'd you think @canihavesomecoffee ?

TheClashster: This is the message posted to the user I think so this should do the job, will fix the typo. Thanks!

mod_ci/controllers.py
canihavesomecoffee: Maybe, but as it won't occur a lot I don't see the need immediately.

canihavesomecoffee: And I'd rather not pollute the db with canceled tests of blocked users.

satyammittal: Github notification is enough, if someone ask, we could see that posting link.

PR #155, opened 2018-01-06
Simplify some lines

canihavesomecoffee:
- Simplify lines that could be put on one line with new max-length
- "true" if foo else "false" type of constructs simplified
- Simplified handling for PR's (merged if blocks together)

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -31,12 +31,10 @@ from mod_auth.controllers import login_required, check_access_rights
 from mod_ci.models import Kvm, MaintenanceMode, BlockedUsers
 from mod_deploy.controllers import request_from_github, is_valid_signature
 from mod_home.models import GeneralData
-from mod_regression.models import Category, RegressionTestOutput, \
-    RegressionTest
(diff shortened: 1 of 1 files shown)


mod_ci/controllers.py
```diff
@@ -133,264 +131,241 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
...
         origin = repo.remote('origin')
     except ValueError:
-        log.critical('[{platform}] Origin remote doesn\'t exist!')
+        log.critical("[{platform}] Origin remote doesn't exist!".format(platform=platform))
         return
+
     fetch_info = origin.fetch()
     if len(fetch_info) == 0:
-        log.warn('[{platform}] No info fetched from remote!')
+        log.info('[{platform}] Fetch from remote returned no new data...'.format(platform=platform))
```
saurabhshri: Probably `"` instead of `'` for consistency.

canihavesomecoffee: That's some nitpicking that I'd rather do once we have nothing else left to correct, but I agree. It's inconsistent over multiple places.

saurabhshri: 😂  👍

PR #157, opened 2018-01-06
[IMPROVEMENT] Add badge

TheFaR8:

- [X] I have used the project briefly.

---

add badge in svg (instead of previous one using svgwrite)


mod_ci/controllers.py
```diff
@@ -622,18 +621,32 @@ def start_ci():
...
-        output = os.path.join(parent_dir, 'static', 'svg', 'status-{platform}.svg'.format(platform=test.platform.value))
-        dwg = svgwrite.Drawing(output, profile='full')
-
+        availableon = os.path.join(parent_dir, 'static', 'svg',
+                                   'status-{platform}.svg'.format(platform=test.platform.value))
+        svglocation = os.path.join(parent_dir, 'static', 'img',
+                                   'status', 'status-{platform}.svg'.format(platform=test.platform.value))
         if status == Status.SUCCESS:
-            dwg.add(dwg.text('Passing', insert=(0, 0.2), fill='green'))
+            svglocation.format(status="SUCCESS", platform=test.platform.value)
```
canihavesomecoffee: This won't work. svglocation has no more variables at this point.

mod_ci/controllers.py
```diff
@@ -622,18 +621,32 @@ def start_ci():
...
+
         elif status == Status.FAILURE:
-            dwg.add(dwg.text('Failing', insert=(0, 0.2), fill='red'))
+            svglocation.format(status="FAILURE", platform=test.platform.value)
+
         else:
-            dwg.add(dwg.text('Unknown', insert=(0, 0.2), fill='gray'))
-        dwg.save()
+            svglocation.format(status="ERROR", platform=test.platform.value)
```
canihavesomecoffee: If you pass in the status into the image, you can just as well supress all branches.

mod_ci/controllers.py
```diff
@@ -765,6 +778,7 @@ def progress_reporter(test_id, token):
                 if status == TestStatus.canceled:
                     state = Status.ERROR
                     message = 'Tests aborted due to an error; please check'
+                    update_build_badge(state, test)
```
canihavesomecoffee: Why was this added?

mod_ci/controllers.py
```diff
@@ -622,19 +621,33 @@ def start_ci():
...
+
         elif status == Status.FAILURE:
-            dwg.add(dwg.text('Failing', insert=(0, 0.2), fill='red'))
-        else:
-            dwg.add(dwg.text('Unknown', insert=(0, 0.2), fill='gray'))
-        dwg.save()
+            availableon.format(status, platform=test.platform.value)
 
+        else:
+            availableon.format(status, platform=test.platform.value)
```
canihavesomecoffee: The same still applies for this location.

Besides this, a .format requires all parameters to be filled in at once, so your first format for availableon will throw a KeyError.

You can do this without any if/else structures within the outer if.

mod_ci/controllers.py
```diff
@@ -622,19 +621,25 @@ def start_ci():
 
 
 def update_build_badge(status, test):
+    '''
+    Build status badge for current test to be displayed on sample-platform.ccextractor.org
+
+    :param status: current testing status
+    :type status: str
+    :param test: current commit that is tested
+    :type test: str
```
canihavesomecoffee: This is incorrect. If `test` was a `str`, then `test.test_type` is not a valid code construct.

mod_ci/controllers.py
```diff
@@ -622,19 +621,25 @@ def start_ci():
...
-        if status == Status.SUCCESS:
-            dwg.add(dwg.text('Passing', insert=(0, 0.2), fill='green'))
-        elif status == Status.FAILURE:
-            dwg.add(dwg.text('Failing', insert=(0, 0.2), fill='red'))
-        else:
-            dwg.add(dwg.text('Unknown', insert=(0, 0.2), fill='gray'))
-        dwg.save()
-
+        availableon = os.path.join(parent_dir, 'static', 'svg',
+                                   '{status}-{platform}.svg'.format(status=Status.value, platform=test.platform.value))
```
canihavesomecoffee: `Status.value` is not correct here. You use the class, but what about the `status` string you passed in as argument?

mod_ci/controllers.py
```diff
@@ -622,19 +621,25 @@ def start_ci():
 
 
 def update_build_badge(status, test):
+    '''
```
canihavesomecoffee: Please use `"""` instead of `'''`, as PEP-257 prescribes.

TheFaR8:
Not so sure about the URL routing part though

canihavesomecoffee:
What are you doing? The previous version was near correct, and you just had a couple of small fixes to do.

This new version makes **no** sense whatsoever.

TheFaR8:
Reverted

mod_ci/controllers.py
```diff
@@ -622,19 +621,27 @@ def start_ci():
...
+
+    :param status: current testing status
+    :type status: str
+    :param test: current commit that is tested
+    :type test: Test
+    :return: Nothing.
+    :rtype: None
+    """
     if test.test_type == TestType.commit:
+        test.status = Status()
```
canihavesomecoffee: Why are you doing this? This makes still no sense. You pass in a `status` as an argument to this method, and then you're not using it?

TheFaR8: change ```test.status``` into ```status``` ?

TheFaR8: oooh
i understand now about your comment above

i misread ah

TheFaR8: changed

thanks :)

canihavesomecoffee:
Please rebase against master, so that the conflicts are gone.

canihavesomecoffee:
Nvm, I'll do it.

PR #158, opened 2018-01-06
[FEATURE] Added BlockedUsers Management

TheClashster:

---

Benevolent dictator don't be mad plis :)


mod_ci/controllers.py
```diff
@@ -903,6 +903,63 @@ def show_maintenance():
...
+@login_required
+@check_access_rights([Role.admin])
+@template_renderer()
+def blocked_users():
+    return {
+        'blocked_users': BlockedUsers.query.order_by(BlockedUsers.userID)
+    }
+
+
+@mod_ci.route('/blocked_users')
```
saurabhshri: Multiple definitions for a single route? You don't need to write function defs here.

`@mod_ci.route('/blocked_users')` for this you'll write exactly what should happen if someone opens this path. Check other snippets in the same file to get an idea.

mod_ci/controllers.py
```diff
@@ -903,6 +905,63 @@ def show_maintenance():
...
+@template_renderer('ci/blocked_users.html')
+def username(u):
+    userID = str(u)
+    api_url = requests.get('https://api.github.com/user/' + userID)
+    userdata = api_url.json()
+    username = userdata['login']
+    return username
+
+
+app.jinja_env.globals.update(username=username)
```
canihavesomecoffee: The app scope is not defined. I'd prefer the filters to be grouped together.

mod_ci/controllers.py
```diff
@@ -43,6 +45,8 @@
 
 mod_ci = Blueprint('ci', __name__)
 
+app = Flask(__name__)
```
canihavesomecoffee: This is a big no-no. There must be only one app defined, and that's in run.py

mod_ci/controllers.py
```diff
@@ -903,6 +907,63 @@ def show_maintenance():
...
+@check_access_rights([Role.admin])
+@template_renderer()
+def blocked_users():
+    return {
+        'blocked_users': BlockedUsers.query.order_by(BlockedUsers.userID)
+    }
+
+
+@mod_ci.route('/blocked_users')
+@login_required
```
canihavesomecoffee: That def, if it's used in a template, doesn't need any of these decorators.

mod_ci/forms.py
```diff
@@ -0,0 +1,16 @@
+from flask_wtf import Form
+from wtforms import IntegerField, SubmitField, StringField
+from wtforms.validators import DataRequired
+
+
+class AddUsersToBlacklist(Form):
+    userID = IntegerField('User ID', [DataRequired(
```
canihavesomecoffee: This can be put onto one line. We allow up to 120 chars :)

mod_ci/forms.py
```diff
@@ -0,0 +1,16 @@
...
+        message='GitHub User ID not filled in')])
+    comment = StringField('Comment')
+    submit = SubmitField('Add User')
+
+
+class RemoveUsersFromBlacklist(Form):
+    userID = IntegerField('User ID', [DataRequired(
```
canihavesomecoffee: Same here

templates/ci/blocked_users.html
```diff
@@ -0,0 +1,108 @@
...
+                    <td>{{ u.comment }}</td>
+                </tr>
+            {% endfor %}
+            </tbody>
+        </table>
+    </div>
+    <div class="row">
+        <h1>Add Users</h1>
+    </div>
+    <form method="post" name="registrationForm" id="registrationForm" action="{{ url_for('.add_user_to_blacklist') }}">
```
canihavesomecoffee: Having duplicate name's and id's is not a good idea. Please correct.

mod_ci/controllers.py
```diff
@@ -903,6 +907,51 @@ def show_maintenance():
...
+    api_url = requests.get('https://api.github.com/user/' + user_id)
+    userdata = api_url.json()
+    username = userdata['login']
+    return username
+
+
+app.jinja_env.globals.update(username=username)
+
+
+def add_user_to_blacklist():
```
canihavesomecoffee: This will need decorators if it is referenced by an url_for.

mod_ci/controllers.py
```diff
@@ -903,6 +907,51 @@ def show_maintenance():
...
+        blocked_user = BlockedUsers(form.userID.data, form.comment.data)
+        if blocked_user == BlockedUsers.query.filter_by(userID=form.userID.data).first():
+            flash('User already blocked.')
+            return redirect(url_for('ci.blocked_users'))
+        else:
+            g.db.add(blocked_user)
+            g.db.commit()
+
+
+def remove_user_from_blacklist():
```
canihavesomecoffee: Same.

mod_ci/controllers.py
TheClashster: Done, correctly I think 🤦‍♂️

templates/ci/blocked_users.html
```diff
@@ -0,0 +1,108 @@
...
+            <div class="medium-6 columns">
+                {{ macros.render_field(form.comment) }}
+            </div>
+            <div class="medium-12 columns">
+                {{ macros.render_field(form.submit) }}
+            </div>
+    <div class="row">
+        <h1>Remove Blocked Users</h1>
+    </div>
+    <form method="post" name="registrationForm" id="registrationForm" action="{{ url_for('.remove_user_from_blacklist') }}">
```
canihavesomecoffee: Same here

mod_ci/controllers.py
canihavesomecoffee: That is in the run.py.

mod_ci/controllers.py
```diff
@@ -903,6 +907,51 @@ def show_maintenance():
...
+app.jinja_env.globals.update(username=username)
+
+
+def add_user_to_blacklist():
+    form = AddUsersToBlacklist()
+    if form.validate_on_submit():
+        blocked_user = BlockedUsers(form.userID.data, form.comment.data)
+        if blocked_user == BlockedUsers.query.filter_by(userID=form.userID.data).first():
+            flash('User already blocked.')
+            return redirect(url_for('ci.blocked_users'))
```
canihavesomecoffee: you can use a `.blocked_users`, as you're in the module.

mod_ci/controllers.py
```diff
@@ -903,6 +907,51 @@ def show_maintenance():
...
+            g.db.commit()
+
+
+def remove_user_from_blacklist():
+    form = RemoveUsersFromBlacklist()
+    if form.validate_on_submit():
+        blocked_user = BlockedUsers.query.filter_by(userID=form.userID.data).first()
+        if blocked_user is None:
+            flash('No such user in Blacklist')
+            return redirect(url_for('ci.blocked_users'))
```
canihavesomecoffee: Same remark as with other redirect.

mod_ci/controllers.py
```diff
@@ -903,6 +907,51 @@ def show_maintenance():
...
+        else:
```
canihavesomecoffee: If you have a return in the if, you don't need this else; just do no else and remove the indendation.

mod_ci/controllers.py
```diff
@@ -903,6 +907,51 @@ def show_maintenance():
...
+
+
+def add_user_to_blacklist():
+    form = AddUsersToBlacklist()
+    if form.validate_on_submit():
+        blocked_user = BlockedUsers(form.userID.data, form.comment.data)
+        if blocked_user == BlockedUsers.query.filter_by(userID=form.userID.data).first():
+            flash('User already blocked.')
+            return redirect(url_for('ci.blocked_users'))
+        else:
```
canihavesomecoffee: Else branch is unnecessary if you have a return in the if.

mod_ci/controllers.py
```diff
@@ -538,15 +539,15 @@ def start_ci():
...
-                user_id = payload['pull_request']['user']['id']
-                if BlockedUsers.query.filter(BlockedUsers.userID == user_id).first() is not None:
-                    g.log.critical("User Blacklisted")
-                    gh_commit.post(
-                        state=Status.ERROR,
-                        description="CI start aborted. You may be blocked from accessing this functionality",
-                        target_url=url_for('home.index', _external=True)
-                    )
-                    return 'ERROR'
+                if BlockedUsers.query.filter(BlockedUsers.userID == payload['pull_request'][
```
canihavesomecoffee: Please undo this change, or rebase against master to have it.

mod_ci/controllers.py
```diff
@@ -866,6 +867,44 @@ def show_maintenance():
...
+@check_access_rights([Role.admin])
+def add_user_to_blacklist():
+    form = AddUsersToBlacklist()
+    if form.validate_on_submit():
+        blocked_user = BlockedUsers(form.userID.data, form.comment.data)
+        if blocked_user == BlockedUsers.query.filter_by(userID=form.userID.data).first():
+            flash('User already blocked.')
+            return redirect(url_for('.blocked_users'))
+        g.db.add(blocked_user)
+        g.db.commit()
```
canihavesomecoffee: You don't handle the case where there are form errors

mod_ci/controllers.py
```diff
@@ -866,6 +867,44 @@ def show_maintenance():
...
+
+
+@mod_ci.route('/blocked_users/add', method=['POST'])
+@login_required
+@check_access_rights([Role.admin])
+def add_user_to_blacklist():
+    form = AddUsersToBlacklist()
+    if form.validate_on_submit():
+        blocked_user = BlockedUsers(form.userID.data, form.comment.data)
+        if blocked_user == BlockedUsers.query.filter_by(userID=form.userID.data).first():
```
canihavesomecoffee: I would just check if there's already an user with that id present (see your check for deleting). Now you compare the returned object's, which can differ on comment. If that happens, you'll get a DB-error below.

mod_ci/controllers.py
```diff
@@ -866,6 +867,44 @@ def show_maintenance():
...
+@check_access_rights([Role.admin])
+def remove_user_from_blacklist():
+    form = RemoveUsersFromBlacklist()
+    if form.validate_on_submit():
+        blocked_user = BlockedUsers.query.filter_by(userID=form.userID.data).first()
+        if blocked_user is None:
+            flash('No such user in Blacklist')
+            return redirect(url_for('.blocked_users'))
+        g.db.remove(blocked_user)
+        g.db.commit()
```
canihavesomecoffee: No handling of form errors ;)

mod_ci/controllers.py
canihavesomecoffee: You forgot to redirect the user, so the user will see a blank page if it all goes fine.

mod_ci/controllers.py
canihavesomecoffee: No redirect if all went fine.

run.py
```diff
@@ -72,6 +73,17 @@ def install_secret_keys(application, secret_session='secret_key', secret_csrf='s
 install_secret_keys(app)
 
 
+def username(u):
+    user_id = str(u)
```
canihavesomecoffee: Is this specific cast required? a str.format below (which requires no casting) might be clearer on intent.

run.py
```diff
@@ -72,6 +73,17 @@ def install_secret_keys(application, secret_session='secret_key', secret_csrf='s
...
+    api_url = requests.get('https://api.github.com/user/' + user_id)
```
canihavesomecoffee: What if this request fails for any reason?

PR #189, opened 2018-02-17
Fixed - [Bug] Foundation upgrade broke lay-out

sanchaymittal:

- [x] I have used the project extensively, but have not contributed previously.

---
Fixed the broken layout, working smoothly in both desktop and mobile layout.

### DIFF
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -334,6 +334,10 @@ section,
 summary {
     display: block;
     padding-bottom: 0.5%;
+    padding-inline-start: 1.5%;
+    position:fixed;
+    bottom:0;
+    background-color: white;	
 }
 
 .row {
@@ -442,4 +446,5 @@ input.toggle-round:checked + label:after {
 
 .flow {
     display: flow-root;
+    padding-bottom: 50px;
 }
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -334,7 +334,9 @@ section,
 summary {
     display: block;
     padding-bottom: 0.5%;
-    padding-inline-start: 1.5%;
+}
+
+footer{
     position:fixed;
     bottom:0;
     background-color: white;
--- a/static/css/app.css
+++ b/static/css/app.css
@@ -337,7 +337,7 @@ summary {
     padding-inline-start: 1.5%;
     position:fixed;
     bottom:0;
-    background-color: white;	
+    background-color: white;
 }
 
 .row {


satyammittal:
@sanchaymittal Please update your branch with master.

static/css/app.css
```diff
@@ -334,6 +334,10 @@ section,
 summary {
     display: block;
     padding-bottom: 0.5%;
+    padding-inline-start: 1.5%;
```
canihavesomecoffee: If you just want to fix the footer, why did you put it in here? Now this will apply to all those elements as well. That makes no sense.

static/css/app.css
```diff
@@ -442,4 +446,5 @@ input.toggle-round:checked + label:after {
 
 .flow {
     display: flow-root;
+    padding-bottom: 50px;
```
canihavesomecoffee: Why is this padding necessary?

sanchaymittal: Padding is required at the end of the page for mobile and other scroll screens.

static/css/app.css
sanchaymittal: I can use a separate id or class for this but the changes now doesn't affect others because of their orientation.

static/css/app.css
sanchaymittal: @canihavesomecoffee ??

static/css/app.css
canihavesomecoffee: Still, I'd like to see it in a separate section.

sanchaymittal: @canihavesomecoffee but it's better this way as it aligned with symmetry with others

satyammittal:
@sanchaymittal Please check the fix on chrome, It looks like
screenshot from 2018-02-21 18-28-23
and One more change, don't add unnecessary tab /+ spaces like you added in 340 line.

sanchaymittal:
In Chrome, I think it's perfectly ok
capture1
capture2
and i haven't noticed the extra tab 
Thanks for that :slightly_smiling_face:

satyammittal:
@sanchaymittal Great, I updated my google chrome to 64.0, Yeah it looks fine in both windows and linux.

static/css/app.css
```diff
@@ -334,6 +334,10 @@ section,
 summary {
     display: block;
     padding-bottom: 0.5%;
+    padding-inline-start: 1.5%;
+    position:fixed;
```
canihavesomecoffee: So any menu element or any other will be fixed to the bottom? I'd still like to see a separate change for footer elements only.

PR #191, opened 2018-03-04
[IMPROVEMENT] added home model testing and small fix in home models

satyammittal:

---
1) added mod_home models unit testing
2) small fix in mod_home.models


tests/test_home/test_models/TestCCExtractorVersion.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertEqual(actual.released, released_date)
+        self.assertEqual(actual.commit, commit)
+
+    def test_that_representation_works(self):
+        version = '1.2.3'
+        released = '2013-02-27T19:35:32Z'
+        commit = '1978060bf7d2edd119736ba3ba88341f3bec3323'
+        actual = CCExtractorVersion(version, released, commit)
+
+        version = '<Version {v}>'.format(v=actual.version)
```
canihavesomecoffee: I'd name this `expected`. That clarifies the intent just a bit more.

tests/test_home/test_models/TestCCExtractorVersion.py
```diff
@@ -0,0 +1,31 @@
...
+            released, '%Y-%m-%dT%H:%M:%SZ').date()
+        commit = '1978060bf7d2edd119736ba3ba88341f3bec3323'
+        actual = CCExtractorVersion(version, released, commit)
+
+        self.assertEqual(actual.version, version)
+        self.assertEqual(actual.released, released_date)
+        self.assertEqual(actual.commit, commit)
+
+    def test_that_representation_works(self):
+        version = '1.2.3'
```
canihavesomecoffee: If you repeat these variables, why not make them class members? :)

tests/test_home/test_models/TestGeneralData.py
```diff
@@ -0,0 +1,25 @@
...
+
+        self.assertEqual(actual.key, key)
+        self.assertEqual(actual.value, value)
+
+    def test_that_representation_works(self):
+        key = 'linux'
+        value = '22'
+        actual = GeneralData(key, value)
+
+        rep = '<GeneralData {key}: {value}>'.format(key=key, value=value)
```
canihavesomecoffee: Same remark: I'd go with `expected`

tests/test_home/test_models/TestCCExtractorVersion.py
```diff
@@ -0,0 +1,31 @@
...
+
+from mock import mock
+
+
+class TestCCExtractorVersion(unittest.TestCase):
+
+    def test_that_init_works_correctly(self):
+        version = '1.2.3'
+        released = '2013-02-27T19:35:32Z'
+        released_date = datetime.strptime(
```
canihavesomecoffee: This can be on one line, no? We're allowing 120 characters...

tests/test_home/TestControllers.py
```diff
@@ -0,0 +1,31 @@
...
+    def setUp(self):
+        with mock.patch('database.create_session') as create_session:
+            create_session.side_effect = mock_database
+            from run import app
+            app.config['TESTING'] = True
+            self.app = app.test_client()
+
+    def test_root(self):
+        response = self.app.get('/')
+        self.assertEqual(response.status_code, 200)
```
canihavesomecoffee: These tests seem pretty useless to me, aside from preventing a potential 500. I'd love to see some actual checks on whether the template is also rendered correctly, but it's a good start.

tests/test_home/TestControllers.py
```diff
@@ -0,0 +1,31 @@
...
+        self.assertEqual(response.status_code, 200)
```
canihavesomecoffee: Duplicate line? :)

tests/test_home/TestControllers.py
```diff
@@ -0,0 +1,31 @@
...
+def mock_database(db_string, drop_tables=False):
+    db = create_session(db_string, drop_tables)
+    mock_db = mock.MagicMock(spec=db)
+    return mock_db
+
+
+class TestControllers(unittest.TestCase):
+
+    def setUp(self):
+        with mock.patch('database.create_session') as create_session:
```
canihavesomecoffee: Instead of mocking the entire database layer, I'd probably initially add an in-memory database and depending on the test, inject some interesting data in it to test the actual rendering of templates.

You can take a look at https://github.com/canihavesomecoffee/ci_workshop for this :+1:

mod_home/models.py
```diff
@@ -35,7 +36,7 @@ def __init__(self, version, released, commit):
         :type commit: str
         """
         self.version = version
-        self.released = datetime.strptime(released, '%Y-%m-%dT%H:%M:%SZ').date
+        self.released = datetime.strptime(released, '%Y-%m-%dT%H:%M:%SZ').date()
```
canihavesomecoffee: Is this a bug you fixed?

satyammittal: yes

tests/test_home/TestControllers.py
satyammittal: ok

.travis.yml
```diff
@@ -5,8 +5,27 @@ python:
   - 3.6
   - nightly
 install:
+  - sudo apt-get install libvirt-dev
```
canihavesomecoffee: Ideally we have an abstraction to prevent this dependency

canihavesomecoffee:
@satyammittal I can't merge this while Travis isn't building properly. It might be more interesting, as I suggested, to make a small abstraction for it to remove the dependency.

thealphadollar:
Hey @satyammittal , I would suggest you to use [git amend](https://www.atlassian.com/git/tutorials/rewriting-history) for appending small changes :smiley:

run.py
```diff
@@ -25,8 +24,7 @@
 app = Flask(__name__)
 app.wsgi_app = ProxyFix(app.wsgi_app)
 # Load config
-config = parse_config('config')
-app.config.from_mapping(config)
+app.config.from_pyfile('config_sample.py')
```
canihavesomecoffee: This is no good ;)

We don't want to always run from config_sample.py.

tests/base.py
```diff
@@ -0,0 +1,51 @@
...
+    @staticmethod
+    def add_ccextractor_version():
+        ccextractorversion = CCExtractorVersion(ccextractor_version.version, ccextractor_version.released,
+                                                ccextractor_version.commit)
+        g.db.add(ccextractorversion)
+        g.db.commit()
+        return
+
+
+class ccextractor_version:
```
canihavesomecoffee: I don't think a class makes a lot of sense for just storing some values in.

I'd either prefer to have them in a variable that is a namedTuple, or just as constants instead.

tests/base.py
```diff
@@ -0,0 +1,51 @@
...
+    version = '1.2.3'
+    released = '2013-02-27T19:35:32Z'
+    commit = '1978060bf7d2edd119736ba3ba88341f3bec3323'
+
+
+class general_data:
```
canihavesomecoffee: Same here. Rather a namedTuple.

tests/base.py
```diff
@@ -0,0 +1,51 @@
...
+        app.config['DATABASE_URI'] = 'sqlite:///:memory:'
+        app.config['WTF_CSRF_ENABLED'] = False
+        app.config['SQLALCHEMY_POOL_SIZE'] = 1
+        return app
+
+    def setUp(self):
+        pass
+
+    @staticmethod
+    def add_last_commit_to_general_data():
```
canihavesomecoffee: If these are called outside the Flask framework, the g.db will not be defined. I'd rather see you pass in the g.db as an argument in the calls.

tests/base.py
```diff
@@ -0,0 +1,51 @@
...
+        generaldata = GeneralData(general_data.key1, general_data.value1)
+        g.db.add(generaldata)
+        g.db.commit()
+        return
+
+    @staticmethod
+    def add_ccextractor_version():
```
canihavesomecoffee: Same here.

run.py
```diff
@@ -69,7 +67,8 @@ def install_secret_keys(application, secret_session='secret_key', secret_csrf='s
         sys.exit(1)
 
 
-install_secret_keys(app)
+if 'TESTING' not in os.environ or os.environ['TESTING'] == 'False':
```
canihavesomecoffee: I'm unsure if this will work, because if the secret keys aren't installed, they default to None in Flask, which is not a good thing (I'm unsure what the CSRF validation will do at that point)

tests/base.py
```diff
@@ -0,0 +1,57 @@
...
+from flask_testing import TestCase
+from database import create_session
+from mod_home.models import GeneralData, CCExtractorVersion
+from collections import namedtuple
+from mock import mock
+
+
+def load_config():
+    return {}
+    pass
```
canihavesomecoffee: No need to pass after a return ;)

tests/base.py
```diff
@@ -0,0 +1,57 @@
...
+        general_data = namedtuple('general_data', 'key value')
+        self.general_data1 = general_data(
+            key='last_commit', value='1978060bf7d2edd119736ba3ba88341f3bec3323')
+        self.general_data2 = general_data(key='linux', value='22')
+        ccextractor_version = namedtuple(
+            'ccextractor_version', 'version released commit')
+        self.ccextractor_version = ccextractor_version(version='1.2.3',
+                                                       released='2013-02-27T19:35:32Z',
+                                                       commit='1978060bf7d2edd119736ba3ba88341f3bec3323')
+        pass
```
canihavesomecoffee: No need for pass

tests/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @staticmethod
+    def add_last_commit_to_general_data(general_data, db):
+        generaldata = GeneralData(general_data.key, general_data.value)
+        db.add(generaldata)
+        db.commit()
+        return
```
canihavesomecoffee: No need for return

tests/base.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @staticmethod
+    def add_ccextractor_version(ccx_ver, db):
+        ccextractorversion = CCExtractorVersion(ccx_ver.version, ccx_ver.released,
+                                                ccx_ver.commit)
+        db.add(ccextractorversion)
+        db.commit()
+        return
```
canihavesomecoffee: No need for a return

PR #194, opened 2018-05-01
[FEATURE]Send Emails to Mailing List

MatejMecka:

---

Based on the Discussions we had on the Slack Chat it would be a good idea since we have a mailing list to send the GitHub issues to the mailing list.

Subject will look like: `GitHub Issue #210`
Body will look like
```
[QUESTION] set subtitle timing from live stream timestamp - jakubvojacek
body of the issue....
```

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -37,6 +37,7 @@ from mod_auth.models import Role
 from mod_home.models import CCExtractorVersion
 from mod_sample.models import Issue
 from mod_test.models import TestType, Test, TestStatus, TestProgress, Fork, TestPlatform, TestResultFile, TestResult
+from mailer import Mailer
 
 if sys.platform.startswith("linux"):
     import libvirt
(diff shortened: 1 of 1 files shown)


mod_ci/controllers.py
```diff
@@ -573,6 +574,16 @@ def start_ci():
             if issue is not None:
                 issue.title = issue_data['title']
                 issue.status = issue_data['state']
+                # Send Email to the Mailing List using the Mailer Module and Mailgun's API
+                issue_id = issue_data['number']
```
canihavesomecoffee: This is in the wrong spot. We want to send an email on every opened issue. If you put it inside this check, it'll only send out emails when the issue is in our DB (which only happens when a user submits the issue through the platform).

You'll have to move up the code ;)

mod_ci/controllers.py
```diff
@@ -570,20 +590,16 @@ def start_ci():
         elif event == "issues":
             issue_data = payload['issue']
             issue = Issue.query.filter(Issue.issue_id == issue_data['number']).first()
+            issue_title = issue_data['title']
+            issue_id = issue_data['number']
+            issue_author = issue_data['user']['login']
+            issue_body = issue_data['body']
+            # Send Email to the Mailing List using the Mailer Module and Mailgun's API
+            inform_mailing_list(issue_id, issue_title, issue_author, issue_body)
```
canihavesomecoffee: This will be sent every time a change is made to the issue. I'd wrap this function call in an if that checks if the issue was opened :)

MatejMecka: Oh shoot. I forgot that

Issue #200, opened 2018-05-20
[BUG] Uploading files through FTP reveals the complete path on the uploads page


canihavesomecoffee:
Sample platform commit (found at the bottom of each page) : **9c40fde5f9**

**In raising this issue, I confirm the following (please check boxes, eg [X]):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/sample-platform/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/sample-platform/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/sample-platform/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.

**My familiarity with the project is as follows (check one, eg [X]):**

- [ ] I have never visited/used the platform.
- [ ] I have used the platform just a couple of times.
- [ ] I have used the platform extensively, but have not contributed previously.
- [x] I am an active contributor to the platform.

---

If a user uploads a file through the website, just the filename is being displayed.

If a user uploads the file through FTP transfer, then the entire path is being displayed. This shouldn't happen, and just have the same behaviour as the upload through the website.

PR #202, opened 2018-06-02
[FIX] fixed issue of full path in queued samples

satyammittal:
#200 

---
Using OS basename, we can find the name of that file and just display that instead of displaying full path. Same thing when storing in QueuedSamples.


run.py
```diff
@@ -100,7 +97,13 @@ def get_github_issue_link(issue_id):
     )
 
 
+def filename(filepath):
```
canihavesomecoffee: If we're not using it, it seems a bit wasteful to add the code.

Issue #24, opened 2017-01-05
Add support for single test runs or smaller sample sets


canihavesomecoffee:
Currently the bot is either testing everything or nothing.

We should add an option on the platform to test single samples, or sub-sets.

Linked to #19.

PR #213, opened 2018-07-07
[FEATURE] Added support for single test runs or smaller sample sets

satyammittal:

---

- User can select subset of tests while running customized test
- admin can change default regression tests by changing active status
- toggle active button added in regression management mode in admin panel
- categories where no test is run will show no tests under them
#24

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -60,10 +60,8 @@ def before_app_request():
     """
     config_entries = get_menu_entries(
         g.user, 'Platform mgmt', 'cog', [], '', [
-            {'title': 'Maintenance', 'icon': 'wrench', 'route':
-                'ci.show_maintenance', 'access': [Role.admin]},
-            {'title': 'Blocked Users', 'icon': 'ban', 'route':
-                'ci.blocked_users', 'access': [Role.admin]}
+            {'title': 'Maintenance', 'icon': 'wrench', 'route': 'ci.show_maintenance', 'access': [Role.admin]},
+            {'title': 'Blocked Users', 'icon': 'ban', 'route': 'ci.blocked_users', 'access': [Role.admin]}
         ]
     )
     if 'config' in g.menu_entries and 'entries' in config_entries:
@@ -136,7 +134,9 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
     Creates testing xml files to test the change in main repo.
     Creates clone with separate branch and merge pr into it.
     """
-    from run import config, log, app
+    from run import config, log, app, get_github_config
+
+    github_config = get_github_config(config)
 
     log.info("[{platform}] Running kvm_processor".format(platform=platform))
     if kvm_name == "":
@@ -205,15 +205,17 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
     finished_tests = db.query(TestProgress.test_id).filter(
         TestProgress.status.in_([TestStatus.canceled, TestStatus.completed])
     ).subquery()
-    fork = Fork.query.filter(Fork.github.like(("%/{owner}/{repo}.git").format(owner=g.github['repository_owner'],
-                                                                              repo=g.github['repository']))).first()
+    fork = Fork.query.filter(Fork.github.like(
+        "%/{owner}/{repo}.git".format(owner=github_config['repository_owner'], repo=github_config['repository'])
+    )).first()
     test = Test.query.filter(
             Test.id.notin_(finished_tests), Test.platform == platform, Test.fork_id == fork.id
     ).order_by(Test.id.asc()).first()
+
     if test is None:
-        test = Test.query.filter(
-                Test.id.notin_(finished_tests), Test.platform == platform
-        ).order_by(Test.id.asc()).first()
+        test = Test.query.filter(Test.id.notin_(finished_tests), Test.platform == platform).order_by(
+            Test.id.asc()).first()
+
     if test is None:
         log.info('[{platform}] No more tests to run, returning'.format(platform=platform))
         return
@@ -258,7 +260,7 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
 
     log.debug("[{p}] We will compare against the results of test {id}".format(p=platform, id=last_commit.id))
 
-    regression_ids = CustomizedTest.get_customized_regression_tests(test.id)
+    regression_ids = test.get_customized_regressiontests()
 
     # Init collection file
     multi_test = etree.Element('multitest')
@@ -272,9 +274,8 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
         check_write = False
         for regression_test in category.regression_tests:
             if regression_test.id not in regression_ids:
-                break
-            else:
-                check_write = True
+                continue
+            check_write = True
             entry = etree.SubElement(single_test, 'entry', id=str(regression_test.id))
             command = etree.SubElement(entry, 'command')
             command.text = regression_test.command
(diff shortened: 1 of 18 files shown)


mod_ci/controllers.py
```diff
@@ -266,7 +269,12 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
         # Create XML file for test
         file_name = '{name}.xml'.format(name=category.name)
         single_test = etree.Element('tests')
+        check_write = False
         for regression_test in category.regression_tests:
+            if regression_test.id not in regression_ids:
+                break
```
canihavesomecoffee: This seems a bit excessive to me.

What happens if you select test 2, 3 if a category holds 1, 2, 3? If my logic isn't misguiding me, it'd break out of the loop for all the remainders.

Continue seems more appropriate here.

mod_ci/controllers.py
```diff
@@ -266,7 +269,12 @@ def kvm_processor(db, kvm_name, platform, repository, delay):
...
+            else:
+                check_write = True
```
canihavesomecoffee: If you do a break/continue on the if condition, there's no need for an else... You can just set `check_write` to True.

mod_customized/controllers.py
```diff
@@ -56,7 +57,10 @@ def index():
         if len(commit_arr) > 0:
             fork_test_form.commit_select.choices = commit_arr
             commit_options = True
+        fork_test_form.regression_test.choices = [(str(regression_test.id), regression_test)
+                                                  for regression_test in RegressionTest.query.all()]
         if fork_test_form.add.data and fork_test_form.validate_on_submit():
+            regression_tests = fork_test_form.regression_test.data
```
canihavesomecoffee: Please put this after the import ;)

mod_customized/controllers.py
```diff
@@ -56,7 +57,10 @@ def index():
         if len(commit_arr) > 0:
             fork_test_form.commit_select.choices = commit_arr
             commit_options = True
+        fork_test_form.regression_test.choices = [(str(regression_test.id), regression_test)
```
canihavesomecoffee: If you properly type the field, you can just pass int values:

https://wtforms.readthedocs.io/en/stable/fields.html#wtforms.fields.SelectField (the coerce)

mod_customized/controllers.py
```diff
@@ -95,6 +104,9 @@ def add_test_to_kvm(username, commit_hash, platforms):
         test = Test(platform, TestType.commit, fork.id, 'master', commit_hash)
         g.db.add(test)
         g.db.commit()
+        for regression_test in regression_tests:
+            customized_test = CustomizedTest(test.id, int(regression_test))
```
canihavesomecoffee: Typing the value of the select will save the conversion here as well.

mod_customized/models.py
```diff
@@ -35,3 +37,36 @@ def __init__(self, user_id, test_id):
...
+        :type test_id: int
+        :param regression_id: The value of the 'regression_id' field of
+         CustomizedTest model
+        :type regression_id: int
+        """
+        self.test_id = test_id
+        self.regression_id = regression_id
+
+    @staticmethod
+    def get_customized_regression_tests(test_id):
```
canihavesomecoffee: This does not really belong here, why not work with a relationship?

mod_regression/controllers.py
```diff
@@ -74,6 +74,22 @@ def test_edit(regression_id):
     pass
 
 
+@mod_regression.route('/test/<regression_id>/toggle')
+def toggle_active_status(regression_id):
+    # Change active status of the regression test
+    regression_test = RegressionTest.query.filter(RegressionTest.id == regression_id).first()
+    if regression_test is None:
+        abort(404)
+    else:
```
canihavesomecoffee: If you do an abort, this else isn't necessary.

templates/custom/index.html
```diff
@@ -32,10 +45,38 @@ <h1>Create custom tests on your fork</h1>
         <div class="medium-6 columns">
             {{ macros.render_field(addTestFork.platform) }}
         </div>
+        <div class="medium-12 columns">
+                <label for="regression_test">Choose Regression Tests</label>
+                {% for category in categories %}
+                <h6 class="category-header" data-category="{{ category.id }}">{{ category.name }} </h6>
+                <div class="hide" data-category="{{ category.id }}">
+                <input type="checkbox" name="{{ category.name }}" onClick="toggle(this)" /> Select All<br/>
```
canihavesomecoffee: Try to bind this at runtime with JQuery instead of hardcoded. Slap a class selector on these checkboxes, and then do a `$('.selector').on('click', toggle);`

templates/custom/index.html
```diff
@@ -52,8 +93,21 @@ <h2>Your tests</h2>
...
+            for(var i=0; i < checkboxes.length; i++)
+            {
+                checkbox = checkboxes[i].querySelectorAll('input')[0];
+                checkbox.checked = source.checked;
+            }
+        }
         $("input[name='commit_select']").change(function(){
             $("#commit_hash").val(this.value)
         });
+        $('.category-header').on('click', function(){
```
canihavesomecoffee: Both this one and the one above aren't wrapped in a `$(document).ready()` call... Which could potentially cause javascript errors. Please do so.

templates/regression/index.html
```diff
@@ -109,5 +112,25 @@ <h5>Regression tests</h5>
             });
             switchCategory();
         });
+        function toggleactive(test_id) {
+            $.ajax({
+                type: "GET",
+                url: '/regression/test/' + test_id + '/toggle',
```
canihavesomecoffee: Try to use the `url_for`, it's available anyhow... Store the url in a variable with a replacement value, and replace that at runtime with the test_id.

templates/regression/index.html
```diff
@@ -109,5 +112,25 @@ <h5>Regression tests</h5>
...
+                contentType: "application/json; charset=utf-8",
+                dataType: "json",
+                beforeSend:function()
+                {
+                    alert("Wait! Status Updating");
```
canihavesomecoffee: Alerts aren't really user-friendly. Can you try to find a neater way?

mod_regression/controllers.py
```diff
@@ -74,6 +74,22 @@ def test_edit(regression_id):
     pass
 
 
+@mod_regression.route('/test/<regression_id>/toggle')
+def toggle_active_status(regression_id):
```
canihavesomecoffee: There's no authentication on this route.

mod_regression/controllers.py
```diff
@@ -74,6 +74,22 @@ def test_edit(regression_id):
...
+    # Change active status of the regression test
+    regression_test = RegressionTest.query.filter(RegressionTest.id == regression_id).first()
+    if regression_test is None:
+        abort(404)
+    else:
+        regression_test.active = 1 ^ regression_test.active
```
canihavesomecoffee: This is "black" magic. Not very readable. Please correct.

mod_regression/controllers.py
```diff
@@ -74,6 +74,22 @@ def test_edit(regression_id):
...
+        g.db.add(regression_test)
```
canihavesomecoffee: This is most likely not necessary to update code.

mod_regression/controllers.py
```diff
@@ -74,6 +73,21 @@ def test_edit(regression_id):
...
 
 
+@mod_regression.route('/test/<regression_id>/toggle')
+@check_access_rights([Role.admin])
+def toggle_active_status(regression_id):
+    # Change active status of the regression test
+    regression_test = RegressionTest.query.filter(RegressionTest.id == regression_id).first()
+    if regression_test is None:
+        abort(404)
+    regression_test.active = 1 - regression_test.active
```
canihavesomecoffee: If it's a boolean, this is still "magic". What's wrong with the `not` keyword?

PR #216, opened 2018-08-03
[FEATURE] Adding Restart / Stop Tests for admin and user

satyammittal:

---

1. Added restart and stop button for admin for any test.
2. Users can restart and stop their tests in customized tests panel.

### DIFF
--- a/templates/test/list_table.html
+++ b/templates/test/list_table.html
@@ -19,7 +19,7 @@
                         <td data-title="Run Errors">{{ "Unknown" if not test.finished else ("Yes" if test.failed else "No") }}</td>
                         <td data-title="Completed">
                             <a href="{{ url_for('test.by_id', test_id=test.id) }}">{{ "Yes" if test.finished else "No" }}</a>
-                            {% if user.is_admin or "{{ url_for(request.endpoint) }}" == "{{ url_for('custom.index'}}" %}
+                            {% if user.is_admin or url_for(request.endpoint) == url_for('custom.index') %}
                                 &nbsp;<a href="{{ url_for('test.restart_test', test_id=test.id) }}" title="Restart Test"><i class="fa fa-refresh"></i></a>
                                 {% if not test.finished %}
                                     &nbsp;<a href="{{ url_for('test.stop_test', test_id=test.id) }}" title="Stop Test"><i class="fa fa-stop"></i></a>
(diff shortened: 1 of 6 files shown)


templates/test/list_table.html
```diff
@@ -19,7 +19,7 @@
                         <td data-title="Run Errors">{{ "Unknown" if not test.finished else ("Yes" if test.failed else "No") }}</td>
                         <td data-title="Completed">
                             <a href="{{ url_for('test.by_id', test_id=test.id) }}">{{ "Yes" if test.finished else "No" }}</a>
-                            {% if user.is_admin or "{{ url_for(request.endpoint) }}" == "{{ url_for('custom.index'}}" %}
+                            {% if user.is_admin or url_for(request.endpoint) == url_for('custom.index') %}
```
canihavesomecoffee: I'm still wondering why you compare using url_for when you can just compare the endpoints?

templates/test/list_table.html
```diff
@@ -17,7 +17,15 @@
                         <td data-title="Details">{{ test.test_type.description }} <a href="{{ test.github_link }}" target="_blank">{{ (test.commit[:7] if test.test_type == TestType.commit else test.pr_nr) }}</a></td>
                         <td data-title="Platform">{{ test.platform.description }}</td>
                         <td data-title="Run Errors">{{ "Unknown" if not test.finished else ("Yes" if test.failed else "No") }}</td>
-                        <td data-title="Completed"><a href="{{ url_for('test.by_id', test_id=test.id) }}">{{ "Yes" if test.finished else "No" }}</a></td>
+                        <td data-title="Completed">
+                            <a href="{{ url_for('test.by_id', test_id=test.id) }}">{{ "Yes" if test.finished else "No" }}</a>
+                            {% if user.is_admin or request.endpoint == 'custom.index' %}
```
canihavesomecoffee: So any user that manager to get this view rendered from the correct place can restart the test... doesn't seem good.

satyammittal: There is a check at backend too, which will check whether the test has been started by the same user or not.
Or other thing we can do is make a function call in test model, which will confirm whether to display start/stop button or not

canihavesomecoffee: Even if there's a back-end check on the restart_test controller, we still want to hide it here whenever it's not applicable. A function in the test model isn't the correct place. Rather call this view with an extra variable that decides whether or not to show it.

PR #247, opened 2018-11-14
Added Validation for wrong commit hash

T1duS:

---

{pull request content here}

### DIFF
--- a/mod_customized/forms.py
+++ b/mod_customized/forms.py
@@ -1,9 +1,11 @@
+from flask import g
 from flask_wtf import FlaskForm
-from wtforms import widgets, StringField, SubmitField, SelectMultipleField, RadioField
+from wtforms import widgets, StringField, SubmitField, SelectMultipleField, RadioField, validators
 from wtforms.validators import DataRequired, url
 from mod_test.models import TestPlatform
 from mod_regression.models import RegressionTest
-
+from mod_auth.controllers import fetch_username_from_token
+import requests
 
 class MultiCheckboxField(SelectMultipleField):
     widget = widgets.ListWidget(prefix_label=False)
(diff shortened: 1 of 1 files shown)


aadibajpai:
@pep8speaks suggest diff

tests/base.py
```diff
@@ -69,7 +69,7 @@ def MockRequests(url, data=None):
                              'created_at': '2011-04-14T16:00:49Z',
                              'state': 'open'}, 201)
     else:
-        return MockResponse({}, 404)
+        return MockResponse({}, 422)
```
satyammittal: revert it back, 404 looks much better as it is default code for not found.

tests/test_customized/TestControllers.py
```diff
@@ -178,3 +178,19 @@ def test_customize_test_creates_with_customize_regression_tests(self, mock_user,
...
+            self.user.name, self.user.email, self.user.password, Role.tester)
+        with self.app.test_client() as c:
+            response = c.post(
+                '/account/login', data=self.create_login_form_data(self.user.email, self.user.password))
+            response = c.post(
+                '/custom/', data=self.create_customize_form('hellofromtheotherside', ['linux'],
+                                                            regression_test=[2]), follow_redirects=True)
+            self.assertEqual(response.status_code, 200)
+            test = Test.query.filter(Test.id == 3).first()
+            self.assertEqual(test,None)
```
saurabhshri: Why is this needed here?

T1duS: To Test that it doesn't create create another customized test.

saurabhshri: Sorry, how does this test that?

T1duS: There are 2 tests added by default in base.py (line 114). If a test would be added, a test of id 3 should be created which means test = Test.query.filter(Test.id == 3).first() would not return None

saurabhshri: 1. In that case shouldn't it be `CustomizedTest` instead of `Test`?
2. What happens if in future we add one or more default tests? Isn't it better to search for the commit hash?
3. Also, if we're just checking if commit hash is being verified fine, I don't think testing that even matters.

T1duS: @saurabhshri so should I remove that part?

PR #252, opened 2018-11-26
[FIX] Regression test delete

T1duS:

---

Fixed regression test delete. It will work once data migration is done.


satyammittal:
@T1duS @saurabhshri we have to fix this first: #254, which will check the constraint in this issue.

T1duS:
Hey @satyammittal @saurabhshri I have both fixed the issue #254 and added unit tests for the bug.
See here - https://github.com/CCExtractor/sample-platform/pull/252/commits/e27d6ba27e3a0edb43be792fb796ef1f3066b40f This is the previous model so it fails the tests whereas https://github.com/CCExtractor/sample-platform/pull/252/commits/120c9dcc267bdbc594bb4a6eb78f3119156874d0 this is the correct model so it passes the tests

tests/base.py
```diff
@@ -164,23 +177,22 @@ def setUp(self):
             TestProgress(2, TestStatus.completed, "Test 2 completed")
         ]
         g.db.add_all(test_result_progress)
+        g.db.commit()
```
saurabhshri: No need to add commit after every addition. You could add this at the end and all additions will be committed.

tests/test_regression/TestControllers.py
```diff
@@ -80,6 +80,34 @@ def test_delete(self):
             )
             self.assertEqual(response.status_code, 302) # 302 is for Redirection
 
+    def test_delete_v2(self):
```
saurabhshri: What does v2 mean?

tests/base.py
satyammittal: I guess committing first is right. If we don't commit it first, then we will get foreign key errors. as we are using "id" while creating a new row of other table.

tests/test_regression/TestControllers.py
satyammittal: i guess he mean version 2. @T1duS change the function name to more meaningful.

tests/base.py
```diff
@@ -101,6 +101,7 @@ def setUp(self):
         self.app.preprocess_request()
         g.db = create_session(
             self.app.config['DATABASE_URI'], drop_tables=True)
+        g.db.execute('pragma foreign_keys=on')
```
satyammittal: @T1duS can you put a comment here for this code. so that we can understand its reason later.

tests/test_regression/TestControllers.py
saurabhshri: @satyammittal Yes, what's version 2? What's version 1? Why we have two versions?

tests/base.py
saurabhshri: @satyammittal Okay, if you've tested it :)

tests/test_regression/TestControllers.py
satyammittal: there are 2 test_delete function, see the function above (line number 55). we can remove the above function. and keep only version 2.

PR #266, opened 2019-05-20
[Fix] Convert Markdown To HTML Email Body

### LINKED ISSUE #264: [BUG] Markdown rendering in email
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

Sample platform commit (found at the bottom of each page) : **X.X.X**

- [ ] I have never visited/used the platform.
- [ ] I have used the platform just a couple of times.
- [ ] I have used the platform extensively, but have not contributed previously.

---

Emails sent to the mailing list when a new bug is opened and so contain markdown that is sent as plaintext instead of rendering it.

thealphadollar:

---

In reference to issue #264 , the PR uses the [HTML method of mailgun API](https://documentation.mailgun.com/en/latest/api-sending.html#sending) to send the issue body (originally in markdown) by first converting it into HTML using [markdown2](https://github.com/trentm/python-markdown2) along with [extras](https://github.com/trentm/python-markdown2/wiki/Extras) to add more features and convert the task list.

Below is an example of how the emails would look after the PR is merged.

OLD

Screenshot from 2019-05-21 00-48-17

NEW

Screenshot from 2019-05-21 00-47-46

The unit-test for inform_mailing_list function has been updated accordingly.

fixes #264


thealphadollar:
cc/ @canihavesomecoffee @satyammittal Please approve :)

aadibajpai:
before/after pics?

thealphadollar:
> before/after pics?

I had added them but were not properly uploaded; rectified now. Please leave your views and suggestions :smile:

thealphadollar:
> Could you add a unittest that tests the new method you added? :)

Will do :)

thealphadollar:
@canihavesomecoffee Test has been added.

tests/test_ci/TestControllers.py
```diff
@@ -183,25 +183,64 @@ def test_customizedtest_added_to_queue(self):
...
+                                    """<li><a target="_blank" href="www.example.com">list_with_link</a></li>\n"""
+                                    "<li>[X] checkbox_list with_code</li>\n"
+                                    "</ul>\n\n"
+                                    "<p><code>code</code></p>\n\n"
+                                    "<blockquote>\n"
+                                    "  <p>quote</p>\n"
+                                    "</blockquote>\n")
+        received_html_issue_body = get_html_issue_body(title, author, body, issue_number, url)
+
+        assert received_html_issue_body == expected_html_issue_body, "wrong issue html [formatted] email returned"
```
canihavesomecoffee: Please don't use assert, rather use the self.assertEqual & other available assert function calls that are available with the testing framework :)

test.txt
```diff
@@ -0,0 +1,57 @@
+CCExtractor version (using the --version parameter preferably) : **0.87**
```
canihavesomecoffee: This file likely shouldn't be in here? :)

tests/test_ci/TestControllers.py
```diff
@@ -183,25 +183,64 @@ def test_customizedtest_added_to_queue(self):
...
+        author = "abcxyz"
+        body = ("**strong_text**<br/>*em_text*\n\n"
+                "- [list_with_link](www.example.com)\n"
+                "- [X] checkbox_list with_code\n\n"
+                "`code`\n\n"
+                "> quote")
+        issue_number = 1
+        url = "www.example.com"
+
+        expected_html_issue_body = ("[BUG] Test Title - abcxyz<br/>\n"
```
canihavesomecoffee: I'd rather see the use of `"""` instead of parentheses

tests/test_ci/TestControllers.py
```diff
@@ -183,25 +183,64 @@ def test_customizedtest_added_to_queue(self):
...
+    @staticmethod
+    def test_get_html_issue_body():
+        """
+        Test the get_html_issue_body for correct email formatting
+        """
+        from mod_ci.controllers import get_html_issue_body
+
+        title = "[BUG] Test Title"
+        author = "abcxyz"
+        body = ("**strong_text**<br/>*em_text*\n\n"
```
canihavesomecoffee: Same here 👍

test.txt
thealphadollar: Removed.

tests/test_ci/TestControllers.py
thealphadollar: The issue with `"""` is that it doesn't account for the identation and hence it looks really bad when we don't want space to be there before the beginning of lines.

Example.
```
def func():
    sentence="""this is first line
this is second line with no space before
                          this is second line with space before but good indentation"""
```
I don't think breaking the indentation will look good.

tests/test_ci/TestControllers.py
thealphadollar: Same reason as above; to give proper indentation and standardized spacing.

tests/test_ci/TestControllers.py
thealphadollar: I would do that, but may I know the reason just out of curiosity? AFAIK using assert makes the tests framework independent and is a better practice (as per the internet) because it allows ease of changing framework.

canihavesomecoffee: I guess the reason is consistency. Although I did a quick search, and it seems that using "assert" is natively supported by other testing frameworks... Let's keep it then this way :)

tests/test_ci/TestControllers.py
canihavesomecoffee: On second thoughts, I think it doesn't make sense to test that the `markdown` dependency is functioning correctly; I'd just test that it gets called :)

That'll also remove the need for this long body ;)

thealphadollar: Done.

tests/test_ci/TestControllers.py
```diff
@@ -183,25 +183,53 @@ def test_customizedtest_added_to_queue(self):
         self.assertNotIn(1, customized_test)
 
     @mock.patch('mailer.Mailer')
-    def test_inform_mailing_list(self, mock_email):
+    @mock.patch('mod_ci.controllers.get_html_issue_body')
+    def test_inform_mailing_list(self, mock_get_html_issue_body, mock_email):
```
canihavesomecoffee: This test doesn't seem to be fully right... The return value of the html_issue_body returns the same content passed to the send_simple_message, while there's a Jinja template parsing in between...

tests/test_ci/TestControllers.py
```diff
@@ -183,25 +183,53 @@ def test_customizedtest_added_to_queue(self):
...
+    @mock.patch('mod_ci.controllers.markdown')
+    def test_get_html_issue_body(mock_markdown):
+        """
+        Test the get_html_issue_body for correct email formatting
+        """
+        from mod_ci.controllers import get_html_issue_body
+
+        title = "[BUG] Test Title"
+        author = "abcxyz"
+        body = ("**strong_text**<br/>*em_text*\n\n"
```
canihavesomecoffee: If we don't care about the contents here, I don't see a reason to have the multiline body :)

tests/test_ci/TestMarkdown.py
```diff
@@ -0,0 +1,128 @@
+from tests.base import BaseTestCase
+
+from markdown2 import markdown
+
+
+class TestMarkdown(BaseTestCase):
+
+    def test_heading(self):
```
canihavesomecoffee: I'd personally split it up into two separate tests. If one fails you see immediately which one :)

thealphadollar: I don't think so; there are two tests just to ensure consistency of the heading tag. Otherwise, we should test till h6, but just to ensure consistency we are going for two. Hence I think it will only make it redundant to write it twice.

tests/test_ci/TestControllers.py
thealphadollar: Jinja template parsing is in html_issue_body which we have mocked here. So, there is no jinja involved in here and we are simply checking whether the function is called and it's return value is properly used or not.

tests/test_ci/TestMarkdown.py
canihavesomecoffee: Also makes sense :)

PR #271, opened 2019-05-25
[IMPROVEMENT] initialize pydocstyle, stylize docstrings in root-level modules

thealphadollar:

---

The Pull Request adds the following features:

- Add `pydocstyle` to the Pipfile.
- Update instructions in the README to use `pydocstyle`
- Add `pydocstylerc` with rules to check only modules and root directory with `numpy` [docstring standards](http://www.pydocstyle.org/en/stable/error_codes.html).
- Add [and update] docstrings in modules present in the root folder.
- Add [and update] docstrings in authentication module.
- Change `cls` to `self` in object method.

### DIFF
--- a/database.py
+++ b/database.py
@@ -89,24 +89,23 @@ class EnumSymbol(object):
 class EnumMeta(type):
     """Generate new DeclEnum classes."""
 
-    def __init__(cls, classname, bases, dict_):
+    def __init__(self, classname, bases, dict_):
         """Initilize EnumMeta with class, name, value and description."""
-        cls._reg = reg = cls._reg.copy()
+        self._reg = reg = self._reg.copy()
         for k, v in dict_.items():
             if isinstance(v, tuple):
-                sym = reg[v[0]] = EnumSymbol(cls, k, *v)
-                setattr(cls, k, sym)
-        return type.__init__(cls, classname, bases, dict_)
+                sym = reg[v[0]] = EnumSymbol(self, k, *v)
+                setattr(self, k, sym)
+        return type.__init__(self, classname, bases, dict_)
 
-    @classmethod
-    def __iter__(cls):
+    def __iter__(self):
         """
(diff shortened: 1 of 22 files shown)


database.py
```diff
@@ -62,76 +61,115 @@ def __init__(self, cls_, name, value, description):
...
-                sym = reg[v[0]] = EnumSymbol(cls, k, *v)
-                setattr(cls, k, sym)
-        return type.__init__(cls, classname, bases, dict_)
+                sym = reg[v[0]] = EnumSymbol(self, k, *v)
+                setattr(self, k, sym)
+        return type.__init__(self, classname, bases, dict_)
 
-    def __iter__(cls):
-        return iter(cls._reg.values())
+    def __iter__(self):
```
canihavesomecoffee: Why change from `cls` to `self`?

run.py
```diff
@@ -88,12 +99,29 @@ def sub_menu_open(menu_entries, active_route):
...
 def date_time_format(value, fmt='%Y-%m-%d %H:%M:%S'):
+    """
+    Add datetime format filter.
+
+    :param value: date
+    :type value: datetime
+    :param fmt: format for the returned string, defaults to '%Y-%m-%d %H:%M:%S'
+    :type fmt: str, optional
+    :return: string representing date-time in given format
+    :rtype: str
```
canihavesomecoffee: Will be be able to drop the type & rtype if we switch to typing?

database.py
thealphadollar: I changed `cls` to `self` since the methods are not class method and hence should have `self` as the first argument.

http://pylint-messages.wikidot.com/messages:e0213

run.py
thealphadollar: Yes, we can drop `type` if we use typing but then if we generate documentation using some service it won't take into account the same and hence having both is redundant but better.

PR #298, opened 2019-06-09
[IMPROVEMENT] Add Setup For Initiating Python 3.7 Compatibility

thealphadollar:

---

The PR concerns #281 and brings in the following changes.

- Update python version to 3.7 in Pipfile
- Add MonkeyType to Pipfile
- Add PyType to Pipfile
- Skip E701 from Pep8 since we will be using the syntax `x: int = 9`
- Refactor README.md with better developer instructions including using PyType and MonkeyType
- Add annotations to all root modules.
- Make Travis Builds on 3.4, 3.5, 3.6 optional to pass since we are switching to 3.7
- Move Travis build dist from trust (default) to xenial since Python 3.7 is not available on trust.
- Move PEP8 checking out of unit-tests
- Move travis installed test requirements to test-requirements.txt

After this I'll be adding unit-tests first and then only adding more annotations.

NOTE: For reasoning on each change, please see the message for the commit that introduces the change.


.travis.yml
```diff
@@ -1,36 +1,32 @@
+dist: xenial
 language: python
 python:
   - 3.4
```
canihavesomecoffee: If we bump the minimal version to 3.7, we should drop testing older versions ;)

.travis.yml
```diff
@@ -1,36 +1,32 @@
...
+  - pip install -r test-requirements.txt
   - pip install -r requirements.txt
-  - pip install Flask-Testing
-  - pip install blinker
 script:
   - dodgy
-  - mypy mod_*
-  - pydocstyle --config=./.pydocstylerc
   - isort -rc --check-only
+  - pydocstyle --config=./.pydocstylerc
```
canihavesomecoffee: Why call this twice with minimal changes in how it's called?

.travis.yml
```diff
@@ -1,36 +1,32 @@
+dist: xenial
 language: python
 python:
   - 3.4
   - 3.5
   - 3.6
-  - nightly
```
aadibajpai: why not keep the nightly too given 3.8 beta is already out

database.py
```diff
@@ -1,11 +1,15 @@
 """handles database session and data-type across app."""
+from __future__ import annotations
```
canihavesomecoffee: I see no use of this import? Can you point out where? PRobably overlooking it.

thealphadollar:
Yeah, thanks for the note.

I'll add it as an optional check.

As discussed with @canihavesomecoffee, we will be focusing mainly on 3.7
and hence only that build will be mandatory to pass.

.travis.yml
thealphadollar: One is py**doc**style and the other is py**code**style

database.py
thealphadollar: https://www.python.org/dev/peps/pep-0563/#id6

This is used to use classes which have not been completely defined but we still want to use them in annotations.

"DeclEnum" L96 function definition is an example of the same.

PR #302, opened 2019-06-17
[IMPROVEMENT] Add and Improve Unit-Tests

thealphadollar:

---

The PR brings the following changes:

- Add unit-tests for mod_deploy and mod_auth.
- Segregate methods used by multiple test files into base.py
- Update to existing changes since they were not triggering `/deploy` properly due to not having set correct environment and headers.
- Remove ununsed code and improve existing methods.

NOTE: Each commit contains a brief bulleted explanation if required.

concerns #284

### DIFF
--- a/mod_auth/forms.py
+++ b/mod_auth/forms.py
@@ -62,20 +62,6 @@ def email_not_in_use(has_user_field=False):
     return _email_not_in_use
 
 
-def role_id_is_valid(form, field):
-    """
-    Check for validity of User's Role.
-
-    :param form: The form which is being passed in
-    :type form: Form
-    :param field : The data value for the 'role' inserted by User
-    :type field : SelectField
-    """
-    role = Role.query.filter(Role.id == field.data).first()
-    if role is None:
-        raise ValidationError('Role id is invalid')
-
-
 class LoginForm(FlaskForm):
     """Render form for User to enter Log in credentials."""
 
--- a/mod_deploy/controllers.py
+++ b/mod_deploy/controllers.py
@@ -88,7 +88,7 @@ def deploy():
 
     x_hub_signature = request.headers.get('X-Hub-Signature')
     # webhook content type should be application/json for request.data to have the payload
-    # request.data is empty in case of x-www-form-urlencoded   
+    # request.data is empty in case of x-www-form-urlencoded
     if not is_valid_signature(x_hub_signature, request.data, g.github['deploy_key']):
         g.log.warning('Deploy signature failed: {sig}'.format(sig=x_hub_signature))
         abort(abort_code)
--- a/mod_deploy/controllers.py
+++ b/mod_deploy/controllers.py
@@ -78,65 +78,62 @@ def is_valid_signature(x_hub_signature, data, private_key):
 def deploy():
     """Deploy the GitHub request to the test platform."""
     from run import app
-    if request.method != 'POST':
-        return 'OK'
-    else:
-        abort_code = 418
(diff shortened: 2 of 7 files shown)


mod_deploy/controllers.py
```diff
@@ -78,65 +78,62 @@ def is_valid_signature(x_hub_signature, data, private_key):
 def deploy():
     """Deploy the GitHub request to the test platform."""
     from run import app
-    if request.method != 'POST':
```
canihavesomecoffee: Why get rid of this check?

thealphadollar: Since `@github_request` checks for the same and hence it is redundant.

thealphadollar: Please resolve conversation if satisfied.

tests/test_deploy/TestControllers.py
```diff
@@ -1,32 +1,228 @@
...
+        headers = generate_git_api_header('push', sig)
+
+        with self.app.test_client() as client:
+            response = client.post(
+                '/deploy', environ_overrides=WSGI_ENVIRONMENT,
+                data=json.dumps(data), headers=headers
+            )
 
         self.assertEqual(response.status_code, 418)
-        self.assertIn("I\\\'m a teapot", str(response.data))
```
canihavesomecoffee: This should stay in :joy:

thealphadollar: Oh, that's the default response we get :laughing: 

I'll let it stay then.

thealphadollar: I don't think we should let it stay since it's a redundant test. The more prominent reason why I don't want it to be there is it misleads the developer (it did me) to believe that it is a response from "our" server and not the default behavior. Hence, I believe only `status_code` should be there.

mod_auth/forms.py
```diff
@@ -62,20 +62,6 @@ def _email_not_in_use(form, field):
     return _email_not_in_use
 
 
-def role_id_is_valid(form, field):
```
canihavesomecoffee: Removed as it's unused?

tests/test_auth/TestForms.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_empty_invalid_password(self):
+        """
+        Test validation fail for zero length password.
+        """
+        pass_field = Field("")
+
+        with self.assertRaises(ValidationError):
+            valid_password(None, pass_field)
+
+    def test_less_than_min_lenght_invalid_password(self):
```
canihavesomecoffee: Typo in `length` ;)

```suggestion
    def test_less_than_min_length_invalid_password(self):
```

tests/test_auth/TestForms.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        Test validation fail for password of length less than min length.
+        """
+        pass_field = Field("".join(['x' * (int(self.app.config['MIN_PWD_LEN']) - 1)]))
+
+        with self.assertRaises(ValidationError):
+            valid_password(None, pass_field)
+
+    def test_more_than_max_lenght_invalid_password(self):
```
canihavesomecoffee: Same

```suggestion
    def test_more_than_max_length_invalid_password(self):
```

mod_auth/forms.py
thealphadollar: Yes. I've mentioned in the commit; there is no model `Role` and hence there is no way we can query it. This is breaking but was never called.

PR #342, opened 2020-01-22
[fix/improvement]fixed the MYSQL condition issue #341 add other checks to the installer 

eshandhawan51:

---

fixed issue #341 and added conditions for not to leave admin email and password empty also removed the "not a valid identifier" warning

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -41,6 +41,9 @@ do
                     exit 1
                 fi
             fi
+			if [  -f "$file" ]; then
+            break
+            fi
         done
     fi
 done
@@ -55,7 +58,7 @@ echo "-------------------------------"
 echo ""
 echo "In order to configure the platform, we need some information from you. Please reply to the following questions:"
 echo ""
-read -e -r -p -s "Password of the 'root' user of MySQL: " -i "" db_root_password
+read -e -r -p  "Password of the 'root' user of MySQL: " -i "" db_root_password
 # Verify password
 supress_warning=$(mysql_config_editor set --login-path=root_login --host=localhost --user=root --password "${db_root_password}") >> "$install_log" 2>&1
 while ! mysql  --login-path=root_login  -e ";" ; do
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -40,6 +40,9 @@ do
                 if [ "$mysql_user_resp" = "Q" ]; then
                     exit 1
                 fi
+            fi
+			if [  -f "$file" ]; then
```
canihavesomecoffee: Your alignment is not ok. Please correct.

install/install.sh
```diff
@@ -55,7 +58,7 @@ echo "-------------------------------"
 echo ""
 echo "In order to configure the platform, we need some information from you. Please reply to the following questions:"
 echo ""
-read -e -r -p -s "Password of the 'root' user of MySQL: " -i "" db_root_password
+read -e -r -p  "Password of the 'root' user of MySQL: " -i "" db_root_password
```
canihavesomecoffee: Why remove the -s? It's needed to not echo the input back, which is on purpose in this case.

install/install.sh
```diff
@@ -154,12 +157,24 @@ echo ""
...
+   echo "Please Re-enter password again ... "
+   read -e -r -p  "Admin password: " admin_password
+   read -e -r -p  "Confirm admin password: " confirm_admin_password
+
+done 
+while [ $admin_password != $confirm_admin_password ]; do
     echo "Entered passwords did not match! Retrying..."
-    read -e -r -p -s "Admin password: " admin_password
-    read -e -r -p -s "Confirm admin password: " confirm_admin_password
+    read -e -r -p "Admin password: " admin_password
```
canihavesomecoffee: Same question here.

install/install.sh
eshandhawan51: -s is the reason for the "not a valid identifier" 
 read: `Password of the 'root' user of MySQL: ': not a valid identifier
which in turn blocks the input

install/install.sh
eshandhawan51: I will create a new PR for the alignment

install/install.sh
eshandhawan51: it has a much worse effect here as it ends up creating a loop of your passwords do not match at this place and blocks the input

install/install.sh
eshandhawan51: sending a new PR with corrections

PR #343, opened 2020-01-22
[fix/improvement]fixed the MYSQL condition issue #341 add other checks to the installer and improvement to  #342

eshandhawan51:

---
improvent to #342

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -41,7 +41,7 @@ do
                     exit 1
                 fi
             fi
-	    if [  -f "$file" ]; then
+            if [  -f "$file" ]; then
                 break
             fi
         done
--- a/install/install.sh
+++ b/install/install.sh
@@ -159,18 +159,14 @@ config_db_uri="mysql+pymysql://${db_user}:${db_user_password}@localhost:3306/${d
 echo ""
 echo "We need some information for the admin account"
 read -e -r -p "Admin username: " -i "admin" admin_name
-read -e -r -p "Admin email: " admin_email
+
 while [$admin_email -z ];do
-   echo "Entered email can't be empty!"
-   echo "Please Re-enter email again ... "
+   echo "Enter Admin email ( It can't be empty! )"
    read -e -r -p "Admin email: " admin_email
 done 
-read -s -e -r -p  "Admin password: " admin_password
-echo ""
-read -s -e -r -p  "Confirm admin password: " confirm_admin_password
+
 while [$admin_password -z ];do
-   echo "Entered password can't be empty!"
-   echo "Please Re-enter password again ... "
+   echo "Enter Admin password (size of password >1) "
    read -s -e -r -p  "Admin password: " admin_password
    echo " "
    read -s -e -r -p  "Confirm admin password: " confirm_admin_password
--- a/install/install.sh
+++ b/install/install.sh
@@ -41,7 +41,7 @@ do
                     exit 1
                 fi
             fi
-		    if [  -f "$file" ]; then
+	    if [  -f "$file" ]; then
                 break
             fi
         done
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -40,6 +40,9 @@ do
                 if [ "$mysql_user_resp" = "Q" ]; then
                     exit 1
                 fi
+            fi
+		    if [  -f "$file" ]; then
```
canihavesomecoffee: This is still not aligned correctly.

install/install.sh
```diff
@@ -55,11 +58,13 @@ echo "-------------------------------"
...
 echo ""
-read -e -r -p -s "Password of the 'root' user of MySQL: " -i "" db_root_password
+read -s -e -r -p  "Password of the 'root' user of MySQL: " -i "" db_root_password
+ echo ""
 # Verify password
+
 supress_warning=$(mysql_config_editor set --login-path=root_login --host=localhost --user=root --password "${db_root_password}") >> "$install_log" 2>&1
 while ! mysql  --login-path=root_login  -e ";" ; do
-    read -e -r -p -s "Invalid password, please retry: " -i "" db_root_password
+    read -s -e -r -p "Invalid password, please retry: " -i "" db_root_password
```
canihavesomecoffee: So swapping around resolves the issue? That's curious :)

install/install.sh
```diff
@@ -154,12 +159,27 @@ echo ""
...
-read -e -r -p -s "Confirm admin password: " confirm_admin_password
-while [ $admin_password -ne $confirm_admin_password ]; do
+while [$admin_email -z ];do
+   echo "Entered email can't be empty!"
+   echo "Please Re-enter email again ... "
+   read -e -r -p "Admin email: " admin_email
+done 
+read -s -e -r -p  "Admin password: " admin_password
+echo ""
+    read -s -e -r -p  "Confirm admin password: " confirm_admin_password
```
canihavesomecoffee: Alignment is not ok

install/install.sh
```diff
@@ -154,12 +159,27 @@ echo ""
...
+while [$admin_password -z ];do
```
canihavesomecoffee: This can be simplified. Why not keep asking for the admin password, and once that's > 1 character proceed to the confirmation?

eshandhawan51: applied the changes to password and email as well

install/install.sh
```diff
@@ -41,6 +41,9 @@ do
                     exit 1
                 fi
             fi
+	    if [  -f "$file" ]; then
```
canihavesomecoffee: This if is still not aligned correctly :(

eshandhawan51: I tried to change it on editor as well as implement change on git hub directly but it isn’t just working :|

canihavesomecoffee:
```suggestion
            if [  -f "$file" ]; then
```

install/install.sh
thealphadollar: @eshandhawan51 What are you trying to solve here? I don't think the ordering in anyway affects the function.

install/install.sh
```diff
@@ -41,6 +41,9 @@ do
                     exit 1
                 fi
             fi
+            if [  -f "$file" ]; then
+                break
+            fi
```
thealphadollar: Why this?

canihavesomecoffee: You obviously didn't follow the Slack discussion; it is because the installer keeps looping otherwise.

install/install.sh
canihavesomecoffee: It did on his distro, so it's a valid fix as far as I'm concerned.

install/install.sh
eshandhawan51: I had an error while installing it that the terminal entered the install mysql condition though it was already installed and ended up forming an infinite loop just printing on the screen "* Installing MySQL (root password will be empty!) " so I added a condition to break the loop from inside if the file is present in case

thealphadollar: Thanks for the update :D

install/install.sh
eshandhawan51: earlier it showed the warning  "': not a valid identifier" in the terminal so I earlier removed the -s from the read function but and the error stopped popping up it also blocked the input through the terminal 
nut later on when I applied it it worked so I didn't remove it

eshandhawan51: technically it shouldn't but somehow it did :)

PR #389, opened 2020-04-14
updated installation guide

zelzhan:

- [X] I have used the project briefly.

---

Update the troubleshooting part, regarding the possible errors on libvirt package.

### DIFF
--- a/install/installation.md
+++ b/install/installation.md
@@ -83,6 +83,13 @@ sudo python bootstrap_gunicorn.py
 * If `gunicorn` boots up successfully, most relevant logs will be stored in
  the `logs` directory. Otherwise they'll likely be in `syslog`.
 
+* If it shows the error regarding the `libvirt`, then there is missing `fftw3.h` file. Try the following:
+```
+sudo apt-get install libvirt-dev
+sudo apt-get install libfftw3-dev
+sudo apt-get install libsndfile1-dev
+```
+
 ## Nginx configuration for X-Accel-Redirect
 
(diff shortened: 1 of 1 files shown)


install/installation.md
```diff
@@ -83,6 +83,13 @@ sudo python bootstrap_gunicorn.py
 * If `gunicorn` boots up successfully, most relevant logs will be stored in
  the `logs` directory. Otherwise they'll likely be in `syslog`.
 
+* If it shows the error regarding the `libvirt`, then there is missing `fftw3.h` file. Try the following:
+```
+sudo apt-get install libvirt-dev
```
thealphadollar, line 88: Can you give an instance when this error occurs? A screenshot would do.

zelzhan: The problem with missing ```fftw3.h``` file occurs on Ubuntu like systems. Have no idea, why this is file is missing out of the box. On windows and CentOS this message does not appear.

About ```sudo apt-get install libvirt-dev```, most of the machines, doesn't have libvirt by default, so I was thinking that this is an essential step which is missed in the installation.

zelzhan: This error appears on installation step

thealphadollar: Sounds good to me. I asked for the details so that if we ever encounter the same, we know where to start.

canihavesomecoffee: Libvirt-bin is installed through the install.sh however, which should (in theory) be enough in combination with the python-libvirt package. I don't mind having the additional note in there though.

PR #479, opened 2020-12-28
Fixed installation error

kvshravan:
**[FIX]**

---

Wrong syntax has been used in /install/install.sh file at [line 87](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L87), which leads to this error :
`./install.sh: line 87: read: Password for admin (will be created): : not a valid identifier
`
And similar error is spotted at lines [96](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L96), [100](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L100), thus not letting user to enter password at lines 96 ,100. This error has been fixed ( -s should be used before -p while using the options for read command) .

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -84,7 +84,7 @@ db_user_exists=$(mysql --login-path=root_login -sse "SELECT EXISTS(SELECT 1 FROM
 
 if [ "${db_user_exists}" = 0 ]; then
     rand_pass=$(< /dev/urandom tr -dc 'a-zA-Z0-9' | fold -w 16 | head -n 1)
-    read -e -r -p -s "Password for ${db_user} (will be created): " -i "${rand_pass}" db_user_password
+    read -e -r -p "Password for ${db_user} (will be created): " -i "${rand_pass}" db_user_password
     # Attempt to create the user
     mysql --login-path=root_login -e "CREATE USER '${db_user}'@'localhost' IDENTIFIED BY '${db_user_password}';" >> "$install_log" 2>&1
     db_user_exists=$(mysql --login-path=root_login -sse "SELECT EXISTS(SELECT 1 FROM mysql.user WHERE user = '$db_user')")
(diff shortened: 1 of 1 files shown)


canihavesomecoffee:
Do you happen to have any reference as to why the `-s` would have to be before the `-p` option?

Aside from that, it looks like the -s is gone in the first change, so that'd have to be placed back. They're all passwords, so they shouldn't be displayed on screen :)

kvshravan:
> Do you happen to have any reference as to why the `-s` would have to be before the `-p` option?
> 
> Aside from that, it looks like the -s is gone in the first change, so that'd have to be placed back. They're all passwords, so they shouldn't be displayed on screen :)

We know -s option masks the password but for the initial time when we randomly make a password , the password should be displayed to the user so for the next time (especially when install fails) if user already exists he can type the password. So the -s option is removed at line 87 and preserved at 96,100.

Secondly , about the -s being used before -p , I've worked it out by writing two simple bash scripts and the one which has -p before -s gives an error. We can also see -s used before -p in lines [61](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L61) and [67](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L67).

kvshravan:
@canihavesomecoffee Wrong syntax has also been used at Lines [163](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L163) and [168](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L168), which gives out unary operator error. The correct syntax would be something like this : 
`while [ -z $admin_email ];do`

canihavesomecoffee:
> @canihavesomecoffee Wrong syntax has also been used at Lines [163](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L163) and [168](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L168), which gives out unary operator error. The correct syntax would be something like this :
> `while [ -z $admin_email ];do`

Feel free to correct those. They worked in the past though, and I haven't tested the installer in a long while... 

Seeing as @thealphadollar added most of those `-s` ones in the (apparently) wrong place, maybe he can check on his system if the behaviour there is different. As far as I can see you are right though, my system doesn't like the `-s` behind the `-p` either.

kvshravan:
> > @canihavesomecoffee Wrong syntax has also been used at Lines [163](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L163) and [168](https://github.com/CCExtractor/sample-platform/blob/master/install/install.sh#L168), which gives out unary operator error. The correct syntax would be something like this :
> > `while [ -z $admin_email ];do`
> 
> Feel free to correct those. They worked in the past though, and I haven't tested the installer in a long while...
> 
> Seeing as @thealphadollar added most of those `-s` ones in the (apparently) wrong place, maybe he can check on his system if the behaviour there is different. As far as I can see you are right though, my system doesn't like the `-s` behind the `-p` either.

Sure , I've updated the PR to fix all syntax errors.

Issue #449, opened 2020-09-08
[Multiple correct outputs] Modify pass/fail decision mechanism


canihavesomecoffee:
The mechanism that decides if a test is passed/failed (https://github.com/CCExtractor/sample-platform/blob/master/mod_test/controllers.py#L130 and further) needs to be modified so that a test passes if the resulting file is one of the approved files.

PR #480, opened 2020-12-30
Modified controller so that a test passes if the resulting file is one of the approved files

kvshravan:
**[IMPROVEMENT]**

---
 #449 

Changed code such that if none of the files match their expected results, the test fails. So if one of the files matches the expected result, the test passes. Correct me if I interpreted this issue wrong and let me know if there is an error regarding the implementation.

### DIFF
--- a/mod_test/controllers.py
+++ b/mod_test/controllers.py
@@ -133,16 +133,16 @@ def get_data_for_test(test, title=None) -> Dict[str, Any]:
             test_error = False
             # A test fails if:
             # - Exit code is not what we expected
-            # - There are result files but one of them is not identical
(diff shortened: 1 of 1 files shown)


canihavesomecoffee:
@kvshravan You should preferably first do the other issues rather than modifying the logic here while the rest is not yet adapted. Even if this logic would be correct, I wouldn't be able to merge this until the rest is done.

kvshravan:
Sorry, I didn't see the issues below. I clicked on the latest issue and was trying to fix it. I even got a doubt when I was working on it ," where are the multiple approved files?" , but just went on doing it. Thanks for giving a clarification.

PR #489, opened 2021-02-09
Multiple correct output Feature 

### LINKED ISSUE #447: [Multiple correct outputs] Modify model for RegressionTestOutput
Modify https://github.com/CCExtractor/sample-platform/blob/master/mod_regression/models.py#L135 so that for a single generated file, multiple correct hashes can be entered. This will likely result in another table with file hashes that link to the id of the RegressionTestOutput.

### LINKED ISSUE #448: [Multiple correct outputs] Update GUI to allow adding a new correct value
Modify the test result view to add a link/button to update the list of correct files.

https://github.com/CCExtractor/sample-platform/blob/master/mod_test/controllers.py
https://github.com/CCExtractor/sample-platform/blob/master/templates/test/by_id.html

### LINKED ISSUE #449: [Multiple correct outputs] Modify pass/fail decision mechanism

kvshravan:
**[FEATURE]** 

---
fixes #447 
fixes #448 
fixes #449 
The following set of features were added :
1. Added multiple correct output feature, the platform can now detect multiple correct output files using RegressionTestOutputFiles table.
2. GUI to add and remove multiple correct outputs in regression test view.
3. Correct output can be chosen from  got values (if they're not None) of last 10 Test Runs.
4. Pass/Fail mechanism has also been modified so that a test is passed when the output file is one of the multiple correct files.

### DIFF
--- a/mod_regression/controllers.py
+++ b/mod_regression/controllers.py
@@ -2,15 +2,20 @@
 
 from flask import (Blueprint, abort, flash, g, jsonify, redirect, request,
                    url_for)
+from sqlalchemy import and_, func
 
 from decorators import template_renderer
 from mod_auth.controllers import check_access_rights, login_required
 from mod_auth.models import Role
-from mod_regression.forms import (AddCategoryForm, AddTestForm,
-                                  ConfirmationForm, EditTestForm)
+from mod_regression.forms import (AddCategoryForm, AddCorrectOutputForm,
+                                  AddTestForm, ConfirmationForm, EditTestForm,
+                                  RemoveCorrectOutputForm)
 from mod_regression.models import (Category, InputType, OutputType,
-                                   RegressionTest, RegressionTestOutput)
+                                   RegressionTest, RegressionTestOutput,
+                                   RegressionTestOutputFiles)
 from mod_sample.models import Sample
+from mod_test.models import (Fork, Test, TestPlatform, TestProgress,
+                             TestResult, TestResultFile, TestStatus, TestType)
 from utility import serve_file_download
 
 mod_regression = Blueprint('regression', __name__)
@@ -205,6 +210,16 @@ def test_result_file(regression_test_output_id):
     return serve_file_download(rto.filename_correct, 'TestResults', 'regression-download')
 
 
+@mod_regression.route('/test/<regression_test_output_id>/download/variant', methods=['GET'])
+def multiple_test_result_file(regression_test_output_id):
+    """View the output files of the regression test (variants)."""
+    rtof = RegressionTestOutputFiles.query.filter(RegressionTestOutputFiles.id == regression_test_output_id).first()
+    if rtof is None:
+        g.log.error(f'requested regression test output file with id: {regression_test_output_id} not found!')
+        abort(404)
+    return serve_file_download(rtof.file_hashes + rtof.output.correct_extension, 'TestResults', 'regression-download')
+
+
 @mod_regression.route('/test/new', methods=['GET', 'POST'])
 @template_renderer()
 @login_required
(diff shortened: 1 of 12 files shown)


mod_regression/controllers.py
```diff
@@ -195,14 +198,22 @@ def toggle_active_status(regression_id):
...
-def test_result_file(regression_test_output_id):
+@mod_regression.route('/test/<regression_test_output_id>/download/<multiple_files>', methods=['GET'])
+def test_result_file(regression_test_output_id, multiple_files):
     """View the output files of the regression test."""
-    rto = RegressionTestOutput.query.filter(RegressionTestOutput.id == regression_test_output_id).first()
-    if rto is None:
-        g.log.error(f'requested regression test output with id: {regression_test_output_id} not found!')
-        abort(404)
-    return serve_file_download(rto.filename_correct, 'TestResults', 'regression-download')
+    if multiple_files == 'True':
```
thealphadollar, line 204: IMO, should we make `multiple_files` an optional field to ensure backward compatibility. Also, I'm not very sure if the exact 'True' should be kept as the parameter.

cc/ @canihavesomecoffee

mod_regression/controllers.py
```diff
@@ -195,14 +198,22 @@ def toggle_active_status(regression_id):
...
+        rtof = RegressionTestOutputFiles.query.filter(RegressionTestOutputFiles.id == regression_test_output_id).first()
+        print(rtof)
```
thealphadollar, line 206: Is this for debugging?

mod_regression/models.py
```diff
@@ -206,3 +208,38 @@ def create_correct_filename(self, name) -> str:
         :rtype: str
         """
         return f"{name}{self.correct_extension}"
+
+
+class RegressionTestOutputFiles(Base):
+    """docstring for RegressionTestOutputFiles."""
```
thealphadollar, line 214: Can we display brief information about the class in the docstring?

thealphadollar (commented):
Hey @kvshravan , thanks for the contribution. I really like how you have used modular code and segregated tasks into different functions.

I've left some comments. I see you have added a few new templates. I'd be interested to see how they look before I approve the PR.

mod_regression/controllers.py
canihavesomecoffee: I'm not so sure why you'd want to keep that optional... It's not like we need to care about broken links for these downloads. As long as all calls inside our program are updated to use it, it should be fine...

canihavesomecoffee (changes requested):
If you could address the things mentioned by @thealphadollar  where I gave a thumbs-up on, it looks good to me to merge. Nice work :+1:

mod_regression/controllers.py
thealphadollar: Yeah. If we are simultaneously updating the calls, then good.

canihavesomecoffee: Well, at least one of the uses is further down in this PR :)

Let's check if there's others.

canihavesomecoffee: @kvshravan Can you make sure that this change doesn't break https://github.com/CCExtractor/sample-platform/blob/master/tests/test_regression/TestControllers.py#L52 (or that this test is adapted)?

kvshravan:
@thealphadollar @canihavesomecoffee  I've segregated the code to download variant files , that should resolve any conflicts for the original code. For your reference I'm attaching two images how the templates look like.
Screenshot from 2021-02-12 13-40-47
Screenshot from 2021-02-12 13-40-40
Screenshot from 2021-02-12 13-40-27

canihavesomecoffee (approved):
LGTM. @thealphadollar is it ok for you?

PR #509, opened 2021-03-07
[FIX] MediaInfo not showing up on the platform

### LINKED ISSUE #507: [BUG] Not all mediainfo is showing up on the site
Sample platform commit (found at the bottom of each page) : 8a90a9cf92

- [] I have never visited/used the platform.
- [] I have used the platform just a couple of times.
- [] I have used the platform extensively, but have not contributed previously.

---

Some recent samples (i.e. https://sampleplatform.ccextractor.org/sample/164) are not showing media info while the xml file is available on the repository. I suspect a parsing issue is at hand. 

The failing xml is attached: [failing.xml.zip](https://github.com/CCExtractor/sample-platform/files/6096148/failing.xml.zip)

kvshravan:

---

Fixes #507

### DIFF
--- a/mod_sample/media_info_parser.py
+++ b/mod_sample/media_info_parser.py
@@ -35,10 +35,10 @@ class MediaInfoFetcher:
                 doc = xmltodict.parse(fd.read())
             except Exception:
                 raise InvalidMediaInfoError("No Mediainfo root element present")
-            if 'Mediainfo' not in doc:
+            if 'MediaInfo' not in doc:
                 raise InvalidMediaInfoError("Mediainfo xml root element not present")
 
-            self.media_info = doc['Mediainfo']
+            self.media_info = doc['MediaInfo']
             self.video_tracks = []      # type: List
             self.caption_tracks = []    # type: List
             self.audio_tracks = []      # type: List
@@ -81,10 +81,10 @@ class MediaInfoFetcher:
             self.general_track = {}
             self.parsed = False
 
-        if 'File' not in self.media_info:
-            raise InvalidMediaInfoError("No File element present in XML")
+        if 'media' not in self.media_info:
+            raise InvalidMediaInfoError("No media element present in XML")
 
-        file_info = self.media_info['File']
+        file_info = self.media_info['media']
         if 'track' not in file_info:
             raise InvalidMediaInfoError("No tracks present in XML")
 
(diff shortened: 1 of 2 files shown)


mod_sample/media_info_parser.py
```diff
@@ -212,13 +215,9 @@ def generate_media_xml(sample) -> MediaInfoFetcher:
...
         process.wait()
         if os.path.isfile(media_info_path):
-            # Load media info, and replace full pathname
+            # Load media info
             tree = etree.parse(media_info_path)
-            for elem in tree.xpath("//track[@type='General']"):
-                for child in elem:
-                    if child.tag == 'Complete_name':
-                        child.text = child.text.replace(media_folder, '')
-                        break
```
canihavesomecoffee, line 221: Why was this deleted? We don't want to leak path names of the file to persons who download the XML.

canihavesomecoffee (changes requested):

canihavesomecoffee:
You'll also have to update the unit tests ;)

mod_sample/media_info_parser.py
kvshravan: There is no such tag as "Complete_name" in the XML generated by the latest version of mediainfo.

mod_sample/media_info_parser.py
```diff
@@ -214,11 +217,13 @@ def generate_media_xml(sample) -> MediaInfoFetcher:
...
+            root = tree.getroot()
+            try:
+                for rootElems in root:
+                    if rootElems.tag.split('}')[1] == 'media':
```
kvshravan, line 223: There is no such tag as `Complete_name` in the XML generated by the latest version of mediainfo, but there is a tag `"media"` which has an attribute `ref` containing full path name of `media_folder`. When looped through the tags , every tag is prefixed with `'{https://mediaarea.net/mediainfo}'`, say for example the tag `media` when printed will look something like this `'{https://mediaarea.net/mediainfo}media'` , the split function is used to split the prefix and tag.

mod_sample/media_info_parser.py
canihavesomecoffee: It's now located in the Media ref attribute. The thing we want to achieve is replacing the full filename (including directories) with just the filename.

kvshravan: Yes, I just made a comment on that in the below part of the code :)

mod_sample/media_info_parser.py
canihavesomecoffee: I wonder if the prefixes can gotten rid of by loading the XSD in some way, but that's not that important to fix.

kvshravan: Yes, [localname](https://stackoverflow.com/questions/18159221/remove-namespace-and-prefix-from-xml-in-python-using-lxml) can be used to eliminate those prefixes.

canihavesomecoffee: That'd look a bit neater I'd say.

Don't forget about the unittests, they're still failing.

kvshravan: Yes I'll update them :)

canihavesomecoffee (approved):

Issue #535, opened 2021-04-27
[BUG] Differences between GitHub comment and actual results


canihavesomecoffee:
Sample platform commit (found at the bottom of each page) : 

**In raising this issue, I confirm the following (please check boxes, eg [X]):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/sample-platform/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm posting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/sample-platform/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/sample-platform/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.

**My familiarity with the project is as follows (check one, eg [X]):**

- [ ] I have never visited/used the platform.
- [ ] I have used the platform just a couple of times.
- [ ] I have used the platform extensively, but have not contributed previously.
- [x] I am an active contributor to the platform.

---

The posted results to a PR on GitHub are inconsistent with what you see when you visit the actual results page.

For example:

https://github.com/CCExtractor/ccextractor/pull/1329#issuecomment-827744099 (table mentions fails, below it says all tests passed)
vs
https://sampleplatform.ccextractor.org/test/3224 (all tests passed)

Another example:
https://github.com/CCExtractor/ccextractor/pull/1329#issuecomment-827495365 (~19 listed failures)
vs
https://sampleplatform.ccextractor.org/test/3225 (5 fails)

pranavrajpal:
I'd like to try working on this.

Is there a way to download the server's database so that I can try and look at the test results more closely? Right now, I can't really tell whether the correct test results are the ones shown on the Sample Platform website, the ones shown in the GitHub comment, or neither of them.

kvshravan:
@pranavrajpal  I don't think you'll be needing the server database, sample database which comes through the installation must be enough, I suppose the issue is with the logic of [comment_pr function](https://github.com/CCExtractor/sample-platform/blob/master/mod_ci/controllers.py#L1148), especially while counting for the passed tests.

canihavesomecoffee:
The correct results are on the page, so it's indeed the commenting function logic which goes wrong for some reason.

[...]

PR #540, opened 2021-05-10
[FIX] Fix incorrect PR comment generation

pranavrajpal:

---

Fixes (part of) #535 (see below for more info)

This fixes a bug in the generation of the PR comment that caused it to mark regression tests as failing if they produced one of the allowed variants of the expected output file, instead of correctly marking these tests as passing. I also added a few tests for this bug.

I don't think this is a complete fix for #535 because I don't think it completely accounts for the difference between https://github.com/CCExtractor/ccextractor/pull/1329#issuecomment-827744099 and https://sampleplatform.ccextractor.org/test/3224. In that comment, the table and the list of failed tests show different results, but the changes I made and the bug I found affect the results for both the table and the list of failed tests in the same way.

Judging from the template for the PR comment, that would probably be caused by `tests` and `failed_tests` in the template getting out of sync, but I can't tell what the cause of that might be just from looking at the code. I also don't know how to debug that problem because I can't tell what regression tests were incorrectly marked as failing.

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -1169,7 +1169,7 @@ class PrCommentInfo:
     failed_tests: List[RegressionTest]
 
 
-def get_info_about_test_for_pr_comment(test_id: int) -> PrCommentInfo:
+def get_info_for_pr_comment(test_id: int) -> PrCommentInfo:
     """Return info about the given test id for use in a PR comment."""
     regression_testid_passed = g.db.query(TestResult.regression_test_id).outerjoin(
         TestResultFile, TestResult.test_id == TestResultFile.test_id).filter(
@@ -1217,7 +1217,7 @@ def comment_pr(test_id, state, pr_nr, platform) -> None:
     """
     from run import app, log
 
-    comment_info = get_info_about_test_for_pr_comment(test_id)
+    comment_info = get_info_for_pr_comment(test_id)
     template = app.jinja_env.get_or_select_template('ci/pr_comment.txt')
     message = template.render(tests=comment_info.category_stats, failed_tests=comment_info.failed_tests,
                               test_id=test_id, state=state, platform=platform)
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -6,9 +6,8 @@ import json
 import os
 import shutil
 import sys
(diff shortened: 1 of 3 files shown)


pranavrajpal:
I don't think the CI failing is related to this PR, seeing as `sudo apt-get install libvirt-dev` seems to be getting a 404 when trying to install those packages.

From what I can tell, this appears to be related to actions/virtual-environments#2104 and there seems to be a relatively simple solution in https://github.com/actions/virtual-environments/issues/2104#issuecomment-731375911. I'll send a PR with that fix.

canihavesomecoffee:
@pranavrajpal can you take a look at the CI? Some tests fail.

pranavrajpal:
I created #543 to fix that bug and one other related bug I found while fixing it, but, for reasons I don't understand, the CI seems to be failing when trying to generate the coverage report. Could you take a look and see if you know what's going wrong?

pranavrajpal:
The CI appears to be passing now. Is there anything else you want me to change/fix?

kvshravan:
Nice work 👍 Can you move the data classes to models.py and then import them as required?

pranavrajpal:
I've moved the data classes to models.py.

I'm slightly confused on how you decide what goes in controllers.py and what goes in models.py. I originally thought models.py was meant for database models and controllers.py was for everything else. Is models.py meant for any classes for storing data, and controllers.py for the rest of the actual logic, or is there some other distinction between what goes in each file?

kvshravan:
I think you can go through mod_sample, afaik Mediainfoparser.py contains custom classes. It's not much of a deal, it's better if the classes aren't defined in controllers.py. @canihavesomecoffee has a say on this

pranavrajpal:
Ok, keeping all of the classes out of controllers.py so that all of the actual logic is kept together makes sense.

The data classes are in models.py and the CI seems to be passing, so let me know if you want me to change anything else.

kvshravan:
LGTM.

canihavesomecoffee:
In general the models were for database (ORM) purposes, but I'd rather see the classes under model than tacked to the controller file. The controller file should be reserved for logic wrt handling requests :)

canihavesomecoffee (approved):

PR #543, opened 2021-05-12
[FIX] Fix tests for complete_signup

pranavrajpal:

---

This contains a few fixes for some tests for `complete_signup` in mod_auth. Those fixes are:
- Fixing a race condition that caused the tests to fail if any of these tests ran more than 1 second after the set up code. More details about the bug can be found at https://github.com/CCExtractor/sample-platform/pull/540#issuecomment-837928237 (fixed in a51ebfecacf0eb3ee966d1c490d708734866bba1)
- Changing `test_if_link_expired` to actually test what would happen when using an expired link. Originally, it was testing what happened when sending an invalid HMAC with a time in the future, which coincidentally resulted in the same outcome. (fixed in cfa6f71a67527949c5219a7ecdc3bfe0e5de6f4c)

I also added some logging to `complete_signup` for when the HMAC or expiry date is invalid to make it easier to debug problems like this if they ever happen again.

### DIFF
--- a/mod_auth/controllers.py
+++ b/mod_auth/controllers.py
@@ -410,7 +410,11 @@ def complete_signup(email: str, expires: int,
                 'expires': expires,
                 'mac': mac
             }
+        else:
+            g.log.error('Invalid HMAC')
 
+    else:
+        g.log.error('HMAC expired')
     flash('The request to complete the registration was invalid. Please enter your email again to start over.',
           'error-message')
     return redirect(url_for('.signup'))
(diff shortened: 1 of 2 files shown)


pranavrajpal:
I don't really understand why the CI is failing. It seems to run all of the tests without failures, and then fails when it's trying to generate the coverage report with the error:

`{'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}`

I'm not very familiar with using Codecov, and I can't seem to find much information about this error online. Does anyone have more information about why this is happening?

thealphadollar (approved):
LGTM.

PR #736, opened 2022-11-09
[IMPROVEMENT] Cancel failing test with build failure

Tarun-Arora:

---

This pull request initiate cancelling of tests if GitHub actions fails along with posting on GitHub.
This pull request allow disallow picking of tests by cron job before completion of their GitHub overflows.

### DIFF
--- a/mod_test/models.py
+++ b/mod_test/models.py
@@ -256,10 +256,7 @@ class Test(Base):
         rtype: list
         """
         customized_test = self.customized_tests
-        if len(customized_test) != 0:
-            regression_ids = [r.regression_id for r in customized_test]
-        else:
-            regression_ids = [r.id for r in RegressionTest.query.all()]
+        regression_ids = [r.regression_id for r in customized_test]
         return regression_ids
 
 
--- a/tests/test_ci/TestControllers.py
+++ b/tests/test_ci/TestControllers.py
@@ -681,7 +681,7 @@ class TestControllers(BaseTestCase):
     @mock.patch('mod_ci.controllers.GeneralData')
     def test_webhook_push_valid(self, mock_gd, mock_github, mock_add_test_entry, mock_request):
         """Test webhook triggered with push event with valid data."""
-        data = {'after': 'abcdefgh'}
+        data = {'after': 'abcdefgh', 'ref': 'refs/heads/master'}
         with self.app.test_client() as c:
             response = c.post(
                 '/start-ci', environ_overrides=WSGI_ENVIRONMENT,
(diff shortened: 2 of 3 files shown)


mod_test/models.py
```diff
@@ -256,10 +256,7 @@ def get_customized_regressiontests(self) -> Any:
         rtype: list
         """
         customized_test = self.customized_tests
-        if len(customized_test) != 0:
-            regression_ids = [r.regression_id for r in customized_test]
-        else:
-            regression_ids = [r.id for r in RegressionTest.query.all()]
+        regression_ids = [r.regression_id for r in customized_test]
```
thealphadollar, line 259: How are we sure that the size of `customized_test` will not be zero?

tests/test_ci/TestControllers.py
```diff
@@ -678,7 +681,7 @@ def test_webhook_push_no_after(self, mock_request):
     @mock.patch('mod_ci.controllers.GeneralData')
     def test_webhook_push_valid(self, mock_gd, mock_github, mock_add_test_entry, mock_request):
         """Test webhook triggered with push event with valid data."""
-        data = {'after': 'abcdefgh'}
+        data = {'after': 'abcdefgh', 'ref': 'refs/heads/master'}
```
thealphadollar, line 684: If the string - `'refs/heads/master'` - is used multiple times, let's move it to a constant string and use it.

thealphadollar (changes requested):
Thanks for the PR, Tarun. I've left a few doubts - let's have a look and then we can merge the changes.

tests/test_ci/TestControllers.py
Tarun-Arora: Just once in this file...

mod_test/models.py
Tarun-Arora: Tests are run only on 2 conditions - upon req. from GH or when admin runs customized test.
1. When the GH workflows are complete `queue_test` function is responsible to add customized test functions
2. Customized test form doesn't allow creating an empty test.

This way, for those tests whose GH Actions are not completed yet, their customized tests would be an empty list, hence we don't have to run their tests right now and wait for it to either have tests or it would get automatically canceled.

thealphadollar: Got it, so in the new flow, it is never possible to have an empty customized_test list.

thealphadollar (approved):

mod_test/models.py
Tarun-Arora: Yes, and we utilising this fact to wait for GH Actions, so that those tests whose GH Actions are pending are not run by the cron job...

PR #810, opened 2023-06-17
[FIX] Update test status, and post to GH, if instance force deleted

Tarun-Arora:

---

This PR is to fix the test status not being updated on the platform and on GitHub and showed "running" in case tests took time longer than the max timeout and the VM instance is deleted as well.

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -17,7 +17,7 @@ import googleapiclient.discovery
 import requests
 from flask import (Blueprint, abort, flash, g, jsonify, redirect, request,
                    url_for)
-from github import Commit, Github, GithubException, Repository
+from github import Commit, Github, GithubException, GithubObject, Repository
 from google.oauth2 import service_account
 from lxml import etree
 from markdown2 import markdown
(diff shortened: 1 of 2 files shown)


Tarun-Arora:
Marked as draft since updating unittests is remaining

mod_ci/controllers.py
```diff
@@ -189,6 +189,21 @@ def delete_expired_instances(compute, max_runtime, project, zone) -> None:
             creationTimestamp = datetime.datetime.strptime(instance['creationTimestamp'], '%Y-%m-%dT%H:%M:%S.%f%z')
             currentTimestamp = datetime.datetime.now(datetime.timezone.utc)
             if currentTimestamp - creationTimestamp >= datetime.timedelta(minutes=max_runtime):
+                # Update test status in database and on GitHub
+                platform_name, test_id = vm_name.split('-')
+                test = Test.query.filter(Test.id == test_id).first()
+                message = "Could not complete test, time limit exceeded"
```
thealphadollar, line 195: Should we add this string as a constant and then use it here?

We can add the constant with other such messaging constants file.

mod_ci/controllers.py
```diff
@@ -650,18 +665,38 @@ def schedule_test(gh_commit: Commit.Commit) -> None:
...
         for platform in TestPlatform:
-            try:
-                gh_commit.create_status(
-                    state=Status.PENDING,
-                    description="Waiting for actions to complete",
-                    context=f"CI - {platform.value}",
-                )
-            except GithubException as a:
-                log.critical(f'Could not post to GitHub! Response: {a.data}')
+            status_description = "Waiting for actions to complete"
```
thealphadollar, line 670: ^ same as above

thealphadollar (approved):
LGTM! Except for a small suggestion.

mod_ci/controllers.py
tarora-eightfold: Actually we should not do this, at least now because:
1. This would have made sense if we could attach a description to the status itself, but since this would somewhat restrict custom messages for the same status, we should allow free text
2. We are not reusing 2 same messages, so even duplication isn't an issue

But definitely if in future it is required, we can make this change...

Tarun-Arora: Actually, we should not do this, at least now, because:

1. This would have made sense if we could attach a description to the status itself, but since this would somewhat restrict custom messages for the same status, we should allow free text
2. We are not reusing two same messages, so even duplication isn't an issue

But definitely a good suggestion. If in future it is required, we can make this change... Thoughts?

Tarun-Arora:
I have tested this PR part-wise; we can now merge this...

canihavesomecoffee:
:+1:

PR #847, opened 2023-08-10
[FIX] Change boolean type from bit to tinyint 

### LINKED ISSUE #812: [BUG][ADMIN] Cannot make a regression test inactive
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

Sample platform commit (found at the bottom of each page) : **X.X.X**

- [ ] I have never visited/used the platform.
- [ ] I have used the platform just a couple of times.
- [ ] I have used the platform extensively, but have not contributed previously.

---

Steps to Reproduce:
1. Go to sample-platform and login as an admin
2. Try to toggle the status of a regression test

Tarun-Arora:

---

Closes #812

### DIFF
--- a/migrations/versions/a5183973c3e9_.py
+++ b/migrations/versions/a5183973c3e9_.py
@@ -16,7 +16,7 @@ depends_on = None
 
 
 def upgrade():
-    op.alter_column('regression_test', 'active', existing_type=sa.Boolean(), type_=sa.Boolean(), nullable=False)
+    op.alter_column('regression_test', 'active', existing_type=sa.dialects.mysql.BIT(1), type_=sa.Boolean(), nullable=False)
 
 
 def downgrade():
--- a/migrations/versions/a5183973c3e9_.py
+++ b/migrations/versions/a5183973c3e9_.py
@@ -5,9 +5,8 @@ Revises: 2e0d2e02a721
 Create Date: 2023-08-10 15:51:13.537000
 
 """
-from alembic import op
 import sqlalchemy as sa
-
+from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = 'a5183973c3e9'
--- a/migrations/versions/a5183973c3e9_.py
+++ b/migrations/versions/a5183973c3e9_.py
@@ -1,0 +1,24 @@
+"""Change SQLAlchemy Boolean bit(1) to bool
+
+Revision ID: a5183973c3e9
+Revises: 2e0d2e02a721
+Create Date: 2023-08-10 15:51:13.537000
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'a5183973c3e9'
+down_revision = '2e0d2e02a721'
+branch_labels = None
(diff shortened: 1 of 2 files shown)


migrations/versions/a5183973c3e9_.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'a5183973c3e9'
+down_revision = '2e0d2e02a721'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column('regression_test', 'active', existing_type=sa.Boolean(), type_=sa.Boolean(), nullable=False)
```
thealphadollar, line 19:
```suggestion
    op.alter_column('regression_test', 'active', existing_type=sa.Boolean(), type_=sa.Boolean(), nullable=False)
```

This seems wrong to me compared to the downgrade step.

thealphadollar (changes requested):

migrations/versions/a5183973c3e9_.py
Tarun-Arora: Can you please explain why it is wrong? (I have tested this on spdev as well...)

thealphadollar: Both the existing type and the type are the same values, "sa.Boolean". I think existing type should be "sa.dialects.mysql.BIT(1)".

Tarun-Arora: Oh yes, actually it is a optional arg, required for MYSQL but it had no effect on the Command generated, it would be something like `ALTER TABLE regression_test MODIFY active bool;`

Tarun-Arora:
> Have you confirmed that after the migration no usage-related changes are required?

Yes

thealphadollar (approved):
LGTM. Thanks for the work, Tarun. Let's merge it and then check if the issue is fixed in production.

Tarun-Arora:
Yes sure, this PR can be merged now

Tarun-Arora:
Just one thing to notice, regression_test with id=1 in the prod db has active as 0 (in bit(1)), but due to this bug it is still being run
After merging of this PR, this regression test would still have active 0 (in tinybit(1)), and now it would not be used in future test runs... So, if we still need that test, admin could toggle it back to True

thealphadollar:
>Just one thing to notice, regression_test with id=1 in the prod db has active as 0 (in bit(1)), but due to this bug it is still being run
After merging of this PR, this regression test would still have active 0 (in tinybit(1)), and now it would not be used in future test runs... So, if we still need that test, admin could toggle it back to True

@Tarun-Arora Let's make the test as active and will disable if need be.

Tarun-Arora:
> > Just one thing to notice, regression_test with id=1 in the prod db has active as 0 (in bit(1)), but due to this bug it is still being run
> > After merging of this PR, this regression test would still have active 0 (in tinybit(1)), and now it would not be used in future test runs... So, if we still need that test, admin could toggle it back to True
> 
> @Tarun-Arora Let's make the test as active and will disable if need be.

Done

PR #926, opened 2025-03-18
[Improvement] Fix first time install errors and update docs

hrideshmg:

- [x] I have used the project extensively, but have not contributed previously.

---

This PR addresses several bugs encountered during the initial setup of the platform on a fresh VPS. The deployment process was _less than pleasant_, and this fix aims to improve the experience by fixing a few of the errors and improving the documentation to add a few missing pieces (like the necessity of `ccextractortester` and webhooks).

## Summary of code changes
- The `sample_db.py` file was missing a few imports which were required for the creation of tables by SQLAlchemy. see [docs](https://flask-sqlalchemy.readthedocs.io/en/stable/models/#defining-models).
- The timestamp in one particular `strptime()` expects millisecond precision, however MySQL [does not enable](https://dev.mysql.com/doc/refman/8.4/en/fractional-seconds.html) this by default. I also noticed that the millisecond format is not used anywhere else in the codebase apart from the initial setup run.
- One division by zero and null value access error.

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -1348,8 +1348,14 @@ def progress_type_request(log, test, test_id, request) -> bool:
 
             for p in times:
                 parts = p.time.split(',')
-                start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S')
-                end = datetime.datetime.strptime(parts[-1], '%Y-%m-%d %H:%M:%S')
+                try:
+                    # Try parsing with microsecond precision first
+                    start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S.%f')
+                    end = datetime.datetime.strptime(parts[-1], '%Y-%m-%d %H:%M:%S.%f')
+                except ValueError:
+                    # Fall back to format without microseconds
+                    start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S')
+                    end = datetime.datetime.strptime(parts[-1], '%Y-%m-%d %H:%M:%S')
                 total_time += int((end - start).total_seconds())
 
             if len(times) != 0:
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -1348,8 +1348,8 @@ def progress_type_request(log, test, test_id, request) -> bool:
 
             for p in times:
                 parts = p.time.split(',')
-                start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S.%f')
-                end = datetime.datetime.strptime(parts[-1], '%Y-%m-%d %H:%M:%S.%f')
+                start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S')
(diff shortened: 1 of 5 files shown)


mod_ci/controllers.py
```diff
@@ -1348,8 +1348,8 @@ def progress_type_request(log, test, test_id, request) -> bool:
 
             for p in times:
                 parts = p.time.split(',')
-                start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S.%f')
-                end = datetime.datetime.strptime(parts[-1], '%Y-%m-%d %H:%M:%S.%f')
+                start = datetime.datetime.strptime(parts[0], '%Y-%m-%d %H:%M:%S')
+                end = datetime.datetime.strptime(parts[-1], '%Y-%m-%d %H:%M:%S')
```
canihavesomecoffee, line 1352: Can you clarify why you changed these lines?

canihavesomecoffee (dismissed):

mod_ci/controllers.py
hrideshmg: MySQL databases by default expect the timestamps to be stored without millisecond precision (the `%f`). The TestProgress model doesn't have any configuration set to enable millisecond precision. I also looked around and noticed that this was the only strptime() call which expected this kind of precision.

hrideshmg:
I've fixed the isort error, the workflow checks should be fine now.

mod_ci/controllers.py
canihavesomecoffee: I'm a bit hesitant to merge this as is because I can't remember (and check right now) if we have the precision in the database or not.

Maybe wrap it in a try-except to catch it just to be sure? Try with precision first and then fallback if it fails?

hrideshmg:
Okay, a bit of a hairy situation here. The test is currently failing due to my time format change, the reason for this has to do with the differences in how SQLite (the DB used for the tests) and MySQL handle dates. 

SQLite does not have a native [DATE](https://www.sqlite.org/lang_datefunc.html) type and just stores it as a UNIX timestamp string which includes fractional seconds, SQLAlchemy when reading this string preserves the fractional seconds.

MySQL however *does* have a native DATE type and since by default it does not enable millisecond precision it truncates the timestamp, so when SQLAlchemy fetches it, there are no fractional seconds.

Here is the output of a small comparison script i wrote:
```python
SQLite timestamp: 2025-03-19 02:08:26.399629
MySQL timestamp: 2025-03-19 02:08:26
```

I can't really think of a perfect solution for this problem, off the top of my head here are a few approaches:
1) Modify the TestProgress model so that it strips the fractional seconds before saving to the DB (this is what I'm leaning towards)
2) Write some kind of middleware to check which DB we are using and truncate if SQLite is detected.
3) Enable millisecond precision on MySQL (don't think this is a good idea since it will change existing stuff on prod)

Any thoughts?

mod_ci/controllers.py
hrideshmg: Yeah, that is probably the safest bet. This also broke tests apparently because SQLite does have precision and the test was reliant on the code being in the current state. I've pushed the fix :+1:

canihavesomecoffee (approved):

PR #983, opened 2025-12-28
fix: Address multiple server errors found in production logs

cfsmp3:
## Summary

This PR fixes multiple server errors identified from analyzing production logs in `/tmp/sp`:

- **Null check in queue_test()** - Prevents crash when Test record is missing
- **CommitStatus subscript fix** - Use attribute access instead of dict syntax for PyGithub objects
- **TempFiles directory creation** - Ensure directories exist before file operations
- **Login redirect fix** - Gracefully handle BuildError when endpoint requires parameters

### 1. `AttributeError: 'NoneType' object has no attribute 'id'` in queue_test
- **Location**: `mod_ci/controllers.py:1305`
- **Cause**: `queue_test()` assumed Test record always exists
- **Fix**: Add null check before accessing `platform_test.id`

### 2. `TypeError: 'CommitStatus' object is not subscriptable`
- **Location**: `mod_ci/controllers.py:1498`
- **Cause**: Code used `status["context"]` but PyGithub returns objects, not dicts
- **Fix**: Changed to `status.context` attribute access

### 3. `FileNotFoundError` in progress_reporter
- **Location**: `mod_ci/controllers.py:2135-2150`
- **Cause**: `TempFiles` and `TestResults` directories may not exist
- **Fix**: Added `os.makedirs(..., exist_ok=True)` before file operations

### 4. `BuildError: Could not build url for endpoint 'sample.download_sample'`
- **Location**: `mod_auth/controllers.py` (login redirect)
- **Cause**: Some endpoints require URL parameters that aren't preserved during login redirect
- **Fix**: Wrap `url_for()` in try/except, redirect to home page on BuildError

## Files Changed
- `mod_ci/controllers.py` - Null check, CommitStatus fix, directory creation
- `mod_auth/controllers.py` - Graceful error handling for login redirects

## Test plan
- [x] Verify login redirect tests pass
- [ ] Verify existing CI tests pass
- [ ] Deploy and monitor logs for reduced 500 errors

### DIFF
--- a/tests/test_ci/test_controllers.py
+++ b/tests/test_ci/test_controllers.py
@@ -1901,7 +1901,10 @@ class TestControllers(BaseTestCase):
 
         mock_log.debug.assert_called_once()
         mock_filename.assert_called_once()
-        self.assertEqual(2, mock_os.path.join.call_count)
+        # 4 calls: temp_dir, temp_path, results_dir, final_path
+        self.assertEqual(4, mock_os.path.join.call_count)
+        # 2 calls: makedirs for TempFiles and TestResults directories
+        self.assertEqual(2, mock_os.makedirs.call_count)
         mock_upload_file.save.assert_called_once()
         mock_open.assert_called_once_with(mock.ANY, "rb")
         mock_os.path.splitext.assert_called_once_with(mock.ANY)


mod_auth/controllers.py
```diff
@@ -59,7 +59,8 @@ def login_required(f: Callable) -> Callable:
     def decorated_function(*args, **kwargs):
         if g.user is None:
             g.log.warning(f'login protected endpoint {request.endpoint} accessed before logging in')
-            return redirect(url_for('auth.login', next=request.endpoint))
+            # Store the full URL path instead of just the endpoint name to preserve URL parameters
+            return redirect(url_for('auth.login', next=request.full_path.rstrip('?')))
```
canihavesomecoffee: If I'm not mistaken this could lead to security risks, are you sure this is safe?

mod_auth/controllers.py
```diff
@@ -225,11 +226,17 @@ def login() -> Union[Response, Dict[str, Union[str, LoginForm]]]:
     """Route for handling the login page."""
     redirect_location = request.args.get('next', '')
 
+    # Validate redirect location to prevent open redirects
+    # Only allow paths starting with / (relative to this site)
+    if redirect_location and (not redirect_location.startswith('/') or redirect_location.startswith('//')):
+        redirect_location = ''
```
canihavesomecoffee: This probably does not catch all cases... So I'd rather see the old version and just handle the error more gracefully.

mod_auth/controllers.py
cfsmp3: Addressed - reverted to using endpoint name and added try/except around url_for() to catch BuildError gracefully. Now redirects to home page when the endpoint requires parameters we don't have.

mod_auth/controllers.py
cfsmp3: Agreed - reverted to the old approach. Now using try/except around url_for() to catch BuildError and redirect to home page instead. This avoids the security concerns while still preventing the 500 errors.

canihavesomecoffee (dismissed):

canihavesomecoffee (approved):

PR #995, opened 2026-01-03
[IMPROVEMENT] Add sample inventory generation to support systematic regression testing

03-SudheshnaReddy:

- [✅ ] I have read and understood the [contributors guide](https://github.com/CCExtractor/sample-platform/blob/master/.github/CONTRIBUTING.md).
- [✅ ] I have checked that another pull request for this purpose does not exist.
- [ ✅] I have considered, and confirmed that this submission will be valuable to others.
- [ ✅] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [✅ ] I give this submission freely, and claim no ownership to its content.

- [✅ ] I have used the project briefly.

---

This PR adds the first building block for systematic regression testing
as discussed recently.

### What this PR does
- Adds a sample inventory utility that inspects samples using ffprobe
  (and CCExtractor output when available)
- Produces structured metadata per sample (container, streams, duration,
  detected caption types)
- Adds unit tests with mocked subprocess output so CI is stable
- Keeps all logic optional and non-fatal when external tools are missing

### Why
This improves visibility into what the current sample set actually covers, which helps reason about regression coverage gaps and select a representative regression suite more systematically.
The intent is to strengthen confidence in the Sample Platform so future refactors (legacy C cleanup, Rust work) can proceed more safely.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -6,6 +6,10 @@ from exceptions import CCExtractorEndedWithNonZero, MissingPathToCCExtractor
 from mod_regression.update_regression import update_expected_results
 from run import app
 
+import json
+from pathlib import Path
+from mod_regression.sample_inventory import inventory_samples
+
 
 @app.cli.command('update')
 @click.argument('path_to_ccex')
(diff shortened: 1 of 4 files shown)


cfsmp3 (changes requested):
The library core (probe_sample/inventory_samples) is reasonable, but:
- B1: the manage.py inventory entry point can't be invoked — code sits after app.cli() (which dispatches + exits), and inventory isn't a registered @app.cli.command. So python manage.py inventory → "No such command". Needs proper click registration.
- B2: confirmed by running — test_inventory_multiple_files(tmp_path) uses a pytest fixture in a nose2 project, so it's silently not collected (nose2 ran 1 of 2 tests). Zero CI coverage while looking tested.
- Plus crude stderr-substring caption detection.

PR #998, opened 2026-01-04
chore(deps): upgrade SQLAlchemy from 1.4.41 to 2.0.45

cfsmp3:
## Summary

This PR upgrades SQLAlchemy to version 2.0.45, addressing the breaking changes from the major version update. This is an equivalent replacement for PR #959 (Dependabot's auto-generated PR) but with the necessary code changes to make all tests pass.

### Changes made:
- **DeclarativeBase update**: Replace deprecated `declarative_base()` with new class-based `DeclarativeBase` pattern
- **Raw SQL execution**: Wrap string SQL with `text()` in `mod_health/controllers.py` and `tests/base.py`
- **DeclEnumType fix**: Use `String(50)` as impl instead of `Enum` to avoid SQLAlchemy 2.0's stricter enum value validation in TypeDecorator
- **Engine creation**: Remove deprecated `convert_unicode` parameter and use `StaticPool` for SQLite in-memory databases to ensure proper connection sharing during tests
- **Type hints**: Use generic `Dialect` type instead of SQLite-specific dialect types

## Test plan
- [x] All 402 tests pass with the upgraded SQLAlchemy version
- [ ] Verify application works correctly in staging environment

### DIFF
--- a/database.py
+++ b/database.py
@@ -190,13 +190,13 @@ class DeclEnumType(TypeDecorator):
         """Get enumeration type of self."""
         return DeclEnumType(self.enum)
 
-    def process_bind_param(self, value: Optional[Any], dialect: Dialect) -> Optional[str]:
+    def process_bind_param(self, value: Optional[EnumSymbol], dialect: Dialect) -> Optional[str]:
         """Get process bind parameter."""
         if value is None:
             return None
         return value.value
 
-    def process_result_value(self, value: Optional[Any], dialect: Dialect) -> Optional[EnumSymbol]:
+    def process_result_value(self, value: Optional[str], dialect: Dialect) -> Optional[EnumSymbol]:
         """Get process result value."""
         if value is None:
             return None


database.py
```diff
@@ -162,32 +175,27 @@ def db_type(cls) -> DeclEnumType:
...
 
-    def process_bind_param(self, value: EnumSymbol, dialect: SQLiteDialect_pysqlite) -> str:
+    def process_bind_param(self, value: Optional[Any], dialect: Dialect) -> Optional[str]:
         """Get process bind parameter."""
         if value is None:
             return None
         return value.value
 
-    def process_result_value(self, value: str, dialect: SQLiteDialect_pysqlite) -> EnumSymbol:
+    def process_result_value(self, value: Optional[Any], dialect: Dialect) -> Optional[EnumSymbol]:
```
canihavesomecoffee: Same here, why replace a str (which value should clearly be) with Any, that even can be null

cfsmp3:
Fixed the type annotations - restored `EnumSymbol` and `str` types instead of using `Any`. The original types were more specific and provide better type safety. Thanks for catching this!

canihavesomecoffee (approved):

PR #999, opened 2026-01-04
fix: suppress AsyncMock RuntimeWarnings in Python 3.13+

cfsmp3:
## Summary

This PR fixes the RuntimeWarnings that appear when running tests with Python 3.13+:
```
RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited
```

### Root cause
In Python 3.13+, `MagicMock` automatically returns `AsyncMock` for method calls that look async-like (such as `commit`, `debug`, `warning`, `error`). When tests mock Flask's `g` object and the code calls `g.db.commit()` or `g.log.debug(...)`, Python warns that these coroutines were never awaited.

### Changes
- Add a `warnings.filterwarnings()` to suppress these specific warnings
- Add `setup_mock_g()` helper function for tests that need explicit MagicMock setup
- Update `create_mock_db_query()` to also set up `g.log` as a MagicMock

## Test plan
- [x] All 402 tests pass
- [x] RuntimeWarnings are no longer displayed (was 48, now 0)

### DIFF
--- a/tests/test_ci/test_controllers.py
+++ b/tests/test_ci/test_controllers.py
@@ -38,6 +38,13 @@ class MockPlatform:
         self.value = 'linux'
 
 
+class MockStatus:
+    """Mock GitHub commit status object."""
+
+    def __init__(self, context):
+        self.context = context
+
+
 class MockFork:
     """Mock fork object."""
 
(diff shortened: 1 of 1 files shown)


canihavesomecoffee:
Please get rid of the SQLAlchemy commit, it still holds changes from the original PR that were changed afterwards...

canihavesomecoffee (dismissed):

PR #1000, opened 2026-01-04
fix: handle log file permission errors gracefully

cfsmp3:
## Summary

Fixes crash on startup when log file is owned by a different user (e.g., root vs www-data).

**Problem:** When the web service runs under `www-data` but the log file `/var/www/sample-platform/logs/platform.log` is owned by `root`, the `LogConfiguration` class would throw a `PermissionError` and crash the entire application on startup.

**Solution:** Handle permission errors gracefully by falling back to console-only logging instead of crashing.

## Changes

- Catch `PermissionError` and `OSError` when creating the file handler
- Fall back to console-only logging instead of crashing
- Print helpful warning message to stderr with fix instructions:
  ```
  [WARNING] Cannot write to log file /path/to/log: Permission denied. 
  Falling back to console-only logging. 
  Fix: sudo chown www-data:www-data /path/to/logs -R
  ```
- Update `file_logger` property to return `Optional[RotatingFileHandler]`
- Update `create_logger` to skip file handler when unavailable
- Added 3 new tests for error handling scenarios

## Test plan

- [x] Existing tests pass
- [x] New tests for PermissionError handling
- [x] New tests for OSError handling
- [x] New tests for create_logger without file logger

### DIFF
--- a/install/deploy/pre_deploy.sh
+++ b/install/deploy/pre_deploy.sh
@@ -101,6 +101,32 @@ if ! git diff --quiet 2>/dev/null; then
     git status --short
 fi
 
+# Check 8: Verify logs directory ownership
+LOGS_DIR="$INSTALL_FOLDER/logs"
+WEB_USER="${WEB_USER:-www-data}"
+if [ -d "$LOGS_DIR" ]; then
+    LOGS_OWNER=$(stat -c '%U' "$LOGS_DIR" 2>/dev/null || echo "unknown")
+    if [ "$LOGS_OWNER" != "$WEB_USER" ]; then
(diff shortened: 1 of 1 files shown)


cfsmp3:
Good point! I've added a check in `pre_deploy.sh` that:

1. Verifies the logs directory has correct ownership (`www-data`)
2. Automatically fixes ownership if running as root (which the deploy script does)
3. **Fails deployment** if ownership cannot be fixed

This ensures the issue is caught and fixed during deployment rather than silently falling back to console-only logging.

The graceful fallback in `log_configuration.py` still serves as a safety net (e.g., if permissions change outside of deployment), but the primary fix is now in the deploy script as you suggested.

canihavesomecoffee (approved):

PR #1023, opened 2026-02-16
[FEATURE] Add Docker Setup and Documentation

pulk17:
[FEATURE]

---

## Docker development environment

Two containers — MySQL 8 and the application under Gunicorn — so a contributor
can get a working platform without installing MySQL, Python and the native
libraries the app needs:

```sh
cp env.example .env      # edit the passwords
docker compose up --build
```

That is the whole setup. The app comes up on <http://localhost:5000> with a
schema, fixture data and a browsable UI.

### Why the entrypoint doesn't just run `flask db upgrade`

It can't — and this is the part worth reviewing.

Replaying the migration chain against an empty database fails:

```
(1091, "Can't DROP 'regression_test_ibfk_1'; check that column/key exists")
```

Migration `2e0d2e02a721` drops foreign keys by the names MySQL generates
automatically. Those names only exist if the schema was first built from the
models, which is what `database.create_session` does on the app's first run.
So the chain assumes a database that was never created by the chain itself.
That is a pre-existing property of the repository, not something Docker
introduced, and it is why nothing here tries to "fix" the migrations.

`install/init_schema.py` therefore branches on whether `alembic_version`
exists:

- **absent** — build the schema from the models and `stamp()` it at head,
  which is what a normal first run does anyway
- **present** — an existing environment or a restored dump, so only apply the
  migrations that are newer

It deliberately never does both: `create_all` would create a table that a
pending migration still expects to create itself, which then fails.

A fresh database is also seeded with the existing `install/sample_db.py`,
because several pages dereference rows they assume are present — the home page
does `GeneralData.query.filter(...).first().value` with no null check, so an
empty schema alone serves a 500 on `/`. Seeding through the repository's own
script keeps that fix out of application code.

### Files

| File | Purpose |
|---|---|
| `Dockerfile` | Runtime image; deps cached ahead of the source copy |
| `docker-compose.yml` | MySQL + app, app waits on the DB healthcheck |
| `docker-entrypoint.sh` | Wait for DB → schema → Gunicorn |
| `config.docker.py` | `config.py` for the container, all values from env |
| `install/wait_for_db.py` | Block until MySQL accepts connections |
| `install/init_schema.py` | Create-or-migrate, described above |
| `install/generate_dev_credentials.py` | Throwaway service account so the GCS client can initialise offline |
| `env.example`, `DOCKER.md`, `.dockerignore` | Configuration and docs |

### One change to application code

`utility.py` — `serve_file_download` unconditionally called
`storage_client_bucket.blob(...)`, which cannot work without a real bucket, so
every download raised in development.

It now serves the file from `SAMPLE_REPOSITORY` when no bucket is configured.
**A configured bucket still takes precedence and production behaviour is
unchanged**, with one exception worth naming: a `GoogleAPIError` from a
genuinely configured bucket now falls through to the local copy and returns
404 if it isn't there, where it previously raised. Happy to narrow that to
only the unconfigured case if you would rather production keep failing loudly.

Nothing else outside the Docker files is touched. The earlier
`mod_sample/controllers.py` edit from review has been dropped.

### Security

- Runs as an unprivileged user (`uid 1001`), not root
- No secret in the repository — `env.example` ships placeholders, and secret
  keys plus dev GCP credentials are generated at image build
- The MySQL port is not published to the host
- Compose refuses to start if `MYSQL_ROOT_PASSWORD` or `MYSQL_PASSWORD` are
  unset rather than defaulting to something guessable
- The root password stays with the database container; the app connects as an
  unprivileged user


mod_sample/controllers.py
```diff
@@ -398,4 +398,4 @@ def delete_sample_additional(sample_id, additional_id):
                 'form': form
             }
         raise SampleNotFoundException(f"Extra file {additional_id} for sample {sample.id} not found")
-    raise SampleNotFoundException(f"Sample with id {sample_id} not found")
+    raise SampleNotFoundException(f"Sample with id {sample_id} not found")
```
canihavesomecoffee: Remove this change please.

.dockerignore
```diff
@@ -0,0 +1,69 @@
...
+temp/
+*.tmp
+
+# Secrets and credentials (prevent accidental inclusion)
+*.pem
+*.key
+service-account.json
+gcp-key.json
+secret_key
+secret_csrf
```
canihavesomecoffee: Add a newline to the end.

docker-compose.yml
```diff
@@ -0,0 +1,64 @@
+services:
+  # --- 1. Database Service ---
```
canihavesomecoffee: Please remove all these unnecessary comments.

docker-entrypoint.sh
```diff
@@ -0,0 +1,292 @@
...
+    'private_key_id': 'docker-dev-key',
+    'private_key': pem,
+    'client_email': 'docker-dev@docker-dev.iam.gserviceaccount.com',
+    'client_id': '000000000000',
+    'auth_uri': 'https://accounts.google.com/o/oauth2/auth',
+    'token_uri': 'https://oauth2.googleapis.com/token',
+}
+with open('$REAL_SA_PATH', 'w') as f:
+    json.dump(sa, f, indent=2)
+"
```
canihavesomecoffee: I'd rather see extra helper files (in the correct subfolder) than almost unreadable python in a shell script.

docker-entrypoint.sh
```diff
@@ -0,0 +1,292 @@
+#!/bin/bash
```
canihavesomecoffee: This file is way too large for a entrypoint of a docker. A lot of these things should be handled at build time (ensuring dirs exist/created etc).

Dockerfile
```diff
@@ -0,0 +1,81 @@
+FROM python:3.11-slim-bullseye
```
canihavesomecoffee: I'd rather see a more recent python version.

env.example
```diff
@@ -0,0 +1,63 @@
+# ============================================================
+#  Sample Platform – Docker Environment Variables
+#  Copy env.example → .env, then edit the values below.
+# ============================================================
+
+# ---------- MySQL ----------
+MYSQL_ROOT_PASSWORD=root
```
canihavesomecoffee: The SP does not need to know the root password for MYSQL

env.example
```diff
@@ -0,0 +1,63 @@
...
+# Application-level DB user (MySQL auto-creates this with access to MYSQL_DATABASE)
+MYSQL_USER=sample_platform
+MYSQL_PASSWORD=sample_platform
+MYSQL_DATABASE=sample_platform
+
+# ---------- Networking ----------
+# Port exposed on the HOST for the Flask app (container always listens on 5000)
+APP_PORT=5000
+# Port exposed on the HOST for direct MySQL access (optional, for debugging)
+DB_EXTERNAL_PORT=3306
```
canihavesomecoffee: This should not be exposed.

canihavesomecoffee (changes requested):

canihavesomecoffee (approved):

PR #1044, opened 2026-03-05
fix(upload): resolve login_required test TODOs & patch Exception TypeError

MrButtCode:
**TL;DR**
1. Fixed a live production bug: `QueuedSampleNotFoundException` was raising a `TypeError` due to a missing mandatory `message` argument across four call sites.
2. Resolved Technical Debt: Fixed and restored two abandoned tests in `test_controllers.py` by resolving import-time decorator mocking failures.

**1. Production Bug Fix** (`exceptions.py`)
**The Problem:** The `QueuedSampleNotFoundException` class was updated at some point to require a mandatory `message` argument in its `__init__`. However, four call sites in `mod_upload/controllers.py` were never updated. If triggered in production, this would crash the application with a `TypeError` rather than raising the catchable exception.
**The Fix:** Added a default value (`message="Queued sample not found"`) to the exception's constructor. This is backwards-compatible: existing call sites without arguments now work safely, while custom messages can still be passed.

**2. Test Restoration** (`test_controllers.py`)
**The Problem:** `test_link_id_confirm_invalid` and `test_link_id_confirm_valid` were commented out. The original author attempted to patch `login_required` using `side_effect=mock_decorator` (which wasn't defined) and fundamentally misunderstood that Python evaluates decorators at import time, making standard runtime `mock.patch` ineffective.
**The Fix:** I implemented two distinct testing strategies to handle the decorator:
- For `test_link_id_confirm_invalid`: Restored the test by defining a module-level `mock_decorator` passthrough and using a `reload()` strategy. By patching the decorator at its source in `mod_auth.controllers` and reloading mod_upload.controllers, the decorators are correctly re-applied under the patch.
- For `test_link_id_confirm_valid`: Rewrote this as a true integration test using the Flask HTTP test client. To satisfy the controller's ownership verification without throwing a 404, I seeded the test database with the complete required relational chain (`User` -> `Upload` -> `Sample` -> `QueuedSample`).

### DIFF
--- a/tests/test_upload/test_controllers.py
+++ b/tests/test_upload/test_controllers.py
@@ -188,7 +188,7 @@ class TestControllers(BaseTestCase):
 
         # Mock g.user to have an 'id' attribute to satisfy the controller
         mock_g.user.id = 1
-        
+
         # Ensure the mocked database queries return an object owned by that user
         mock_queue.query.filter.return_value.first.return_value.user_id = mock_g.user.id
         mock_sample.query.filter.return_value.first.return_value.upload.user_id = mock_g.user.id
--- a/exceptions.py
+++ b/exceptions.py
@@ -4,7 +4,7 @@ import sys
 
 class QueuedSampleNotFoundException(Exception):
     """Raised when a queued sample cannot be found by the given ID."""
-    
+
     def __init__(self, message="Queued sample not found."):
         # By providing a default, all existing raise sites that call
         # this exception without arguments will still work, and new


tests/test_upload/test_controllers.py
```diff
@@ -170,35 +184,64 @@ def test_upload_ftp_forbidden_extension(self, mock_shutil, mock_hash, mock_magic
...
-
-    #     self.assertEqual(response, mock_redirect())
-    #     mock_queue.query.filter.assert_called_once_with(mock_queue.id == 1)
-    #     mock_sample.query.filter.assert_called_once_with(mock_sample.id == 1)
+    @mock.patch('mod_auth.controllers.login_required', side_effect=mock_decorator)
+    def test_link_id_confirm_invalid(self, mock_login):
+        """Confirm that an unrecognised upload/sample pair raises QueuedSampleNotFoundException."""
+        # Patching the source module and reloading the consumer forces all
+        # @login_required decorators in mod_upload.controllers to be re-applied
+        # under the mock, since decorators are evaluated at import time.
```
canihavesomecoffee: Why would you add this comment?

tests/test_upload/test_controllers.py
```diff
@@ -170,35 +184,64 @@ def test_upload_ftp_forbidden_extension(self, mock_shutil, mock_hash, mock_magic
...
+        db_user = User.query.filter(User.email == self.user.email).first()
+
+        with self.app.test_client() as c:
+            c.post('/account/login', data=self.create_login_form_data(
+                self.user.email, self.user.password))
+            
+            filehash = 'confirm_test_hash'
+            
+            # The controller checks sample.upload.user_id == g.user.id, which means
+            # Sample and Upload must exist and be correctly linked before the request.
```
canihavesomecoffee: Same here, and on other places.

MrButtCode: You're right, they were unnecessarily verbose. I've removed them.

PR #1059, opened 2026-03-11
[Feature]: Added artifacts for ai

pulk17:
[FEATURE] AI-Accessible Test Artifacts & Structured Test Report Endpoint

---

## What does this PR do?

This PR introduces artifact capture and upload from CI test VMs to GCS, along with a set of new download endpoints and a structured JSON report endpoint designed to make test results fully accessible — including to AI agents.

A demo video is attached below showcasing all the new features end to end.

https://github.com/user-attachments/assets/34dadbc9-4ec9-4f1b-ac89-fa3aef8d9f73

---

## Objective

Right now, when a test run fails, debugging is hard. You can see that something went wrong, but you can't easily grab the binary that was tested, see what it actually printed, or get the coredump if it crashed. Everything lives ephemerally on the VM and disappears. This PR fixes that by persisting the important artifacts to GCS and making them accessible via clean download URLs.

The `ai.json` endpoint was also something I wanted to add to make it possible for AI agents (or any automated tooling) to consume a full structured report of a test run — with direct download links to every artifact — without needing to scrape HTML.

---

### `install/ci-vm/ci-linux/ci/runCI`

- Added a **wrapper script** that intercepts every ccextractor invocation during testing, tees all stdout/stderr into a combined log at `/tmp/combined_stdout.log`, and preserves the original exit code faithfully. This means you get a single unified log of everything ccextractor printed across all test cases in a run.
- After tests complete, the following artifacts are uploaded to GCS under `test_artifacts/{test_id}/`:
  - The `ccextractor` binary itself
  - The combined stdout/stderr log
  - The first coredump found in `/tmp/coredumps/`, if one was produced
- `testID` is now passed as VM instance metadata so the CI script knows where to namespace the uploads.

### `mod_ci/controllers.py`

- `testID` is now included in the metadata passed to both Linux and Windows VM instances at creation time.
- Fixed two instances of `os.rename()` → `os.replace()` in upload handlers. `os.rename` can fail silently or raise an error when source and destination are on different filesystems (which can happen with temp directories). `os.replace` handles this correctly.

### `mod_test/controllers.py`

- Refactored `download_build_log_file` to use `send_from_directory` directly instead of going through `serve_file_download`, which was doing unnecessary GCS signed URL generation for a file that's already on disk.
- Added a shared `_artifact_redirect` helper that looks up a blob in GCS, generates a short-lived signed URL, and redirects to it — or returns 404 if the blob doesn't exist. All new download endpoints use this.
- Added the following new routes:
  - `/<test_id>/binary` — downloads the ccextractor binary used in the test (tries Linux name first, then Windows)
  - `/<test_id>/coredump` — downloads the coredump if one was captured
  - `/<test_id>/combined-stdout` — downloads the unified stdout/stderr log
  - `/<test_id>/regression/<regression_test_id>/<output_id>/output-got` — downloads the actual output file for a specific test case
  - `/<test_id>/regression/<regression_test_id>/<output_id>/output-expected` — downloads the expected output file for a specific test case
  - `/<test_id>/sample/<sample_id>` — downloads the sample file used in a regression test
  - `/<test_id>/ai.json` — returns a full structured JSON report of the test run (see below)

### `utility.py`

- Fixed a path construction bug where `os.path.join` was used to build GCS blob paths. Since GCS paths are always forward-slash separated and `os.path.join` behaves differently on Windows, this is now done with a manual `/`.join` with empty-part filtering.
- Fixed the default value for `GCS_SIGNED_URL_EXPIRY_LIMIT` from `''` (an empty string that would crash `timedelta`) to `30`.

---

## The `ai.json` Endpoint

`GET /<test_id>/ai.json` returns a JSON object that looks roughly like:
```json
{
  "test_id": 123,
  "commit": "abc123",
  "platform": "linux",
  "branch": "master",
  "status": "completed",
  "binary_url": "https://...",
  "coredump_url": null,
  "log_url": "https://...",
  "combined_stdout_url": "https://...",
  "summary": { "total": 50, "passed": 48, "failed": 2 },
  "test_cases": [
    {
      "regression_test_id": 7,
      "category": "...",
      "sample_filename": "sample.ts",
      "sample_url": "https://...",
      "arguments": "-out=ttxt",
      "result": "Fail",
      "exit_code": 1,
      "expected_exit_code": 0,
      "runtime_ms": 4231,
      "how_to_reproduce": "./ccextractor -out=ttxt sample.ts",
      "outputs": [
        {
          "output_id": 3,
          "correct_extension": ".ttxt",
          "expected_url": "https://...",
          "got_url": "https://...",
          "diff_url": "https://..."
        }
      ]
    }
  ],
  "how_to_reproduce": "Download the binary and sample, then run: ./ccextractor {arguments} {sample_filename}"
}
```

Every URL in the response is a signed GCS download link valid for 30 minutes. The intent is that this endpoint gives an AI agent (or any automated system) everything it needs to fully investigate a test failure — the binary, the sample, the expected and actual outputs, the diff, and the full stdout log — without any additional context.

---

## Notes

- The new artifact download routes do not require authentication. This is intentional — the signed URLs are short-lived and the artifacts contain no sensitive data, and requiring auth would prevent AI agents from being able to use them directly. 
- The wrapper script uses a hardcoded `/tmp/combined_stdout.log` path, which is fine given each test run gets its own VM.

### DIFF
--- a/install/ci-vm/ci-linux/ci/runCI
+++ b/install/ci-vm/ci-linux/ci/runCI
@@ -7,6 +7,11 @@
 
 DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
+# Enable coredump capture
+ulimit -c unlimited
+mkdir -p /tmp/coredumps
+echo "/tmp/coredumps/core.%e.%p" | sudo tee /proc/sys/kernel/core_pattern > /dev/null
+
 if [ ! -f "$DIR/variables" ]; then
     # No variable file defined
     sudo shutdown -h now
@@ -123,10 +128,69 @@ if [ -e "${dstDir}/ccextractor" ]; then
         echo "=== CCExtractor Binary Version ===" >> "${logFile}"
         ./ccextractor --version >> "${logFile}" 2>&1
         echo "=== End Version Info ===" >> "${logFile}"
-        postStatus "testing" "Running tests"
+
+
+        ccextractor_path="$(pwd)/ccextractor"
+        combined_stdout="/tmp/combined_stdout.log"
+        : > "${combined_stdout}"
+
+        # Create a wrapper script that tees stdout/stderr to a combined log
+        wrapper_path="$(pwd)/ccextractor_wrapper"
+        cat > "${wrapper_path}" << 'WRAPPER_EOF'
+#!/bin/bash
+COMBINED_LOG="/tmp/combined_stdout.log"
+REAL_BINARY="PLACEHOLDER_BINARY"
+EXIT_CODE_FILE="/tmp/.wrapper_exit_code"
+echo "=== TEST INVOCATION: $@ ===" >> "$COMBINED_LOG"
+{ "$REAL_BINARY" "$@" 2>&1; echo $? > "$EXIT_CODE_FILE"; } | tee -a "$COMBINED_LOG"
+exit_code=$(cat "$EXIT_CODE_FILE")
+echo "=== EXIT CODE: ${exit_code} ===" >> "$COMBINED_LOG"
+echo "" >> "$COMBINED_LOG"
+exit $exit_code
+WRAPPER_EOF
+        sed -i "s|PLACEHOLDER_BINARY|${ccextractor_path}|" "${wrapper_path}"
+        chmod +x "${wrapper_path}"
+
         executeCommand cd ${suiteDstDir}
-        executeCommand ${tester} --debug --entries "${testFile}" --executable "ccextractor" --tempfolder "${tempFolder}" --timeout 600 --reportfolder "${reportFolder}" --resultfolder "${resultFolder}" --samplefolder "${sampleFolder}" --method Server --url "${reportURL}"
+        executeCommand ${tester} --debug --entries "${testFile}" --executable "${wrapper_path}" --tempfolder "${tempFolder}" --timeout 600 --reportfolder "${reportFolder}" --resultfolder "${resultFolder}" --samplefolder "${sampleFolder}" --method Server --url "${reportURL}"
+
+        # Upload AI artifacts to GCS
+        gcs_bucket=$(curl -s "http://metadata/computeMetadata/v1/instance/attributes/bucket" -H "Metadata-Flavor: Google")
+        test_id=$(curl -s "http://metadata/computeMetadata/v1/instance/attributes/testID" -H "Metadata-Flavor: Google")
+        token=$(curl -s "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google" | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])")
+
+        upload_artifact() {
+            local file_path="$1"
+            local dest_path="$2"
+            if [ -f "$file_path" ]; then
+                local http_code
+                http_code=$(curl -s -X POST --data-binary @"$file_path" \
+                    -H "Authorization: Bearer $token" \
+                    -H "Content-Type: application/octet-stream" \
+                    -w "%{http_code}" \
+                    -o /dev/null \
+                    "https://storage.googleapis.com/upload/storage/v1/b/${gcs_bucket}/o?uploadType=media&name=${dest_path}")
+                if [ -z "$http_code" ] || [ "$http_code" -ne 200 ]; then
+                    echo "GCS upload failed for ${dest_path}: HTTP ${http_code:-no_response}" >> "${logFile}"
+                fi
+            fi
+        }
+
+        upload_artifact "$ccextractor_path" "test_artifacts/${test_id}/ccextractor"
+
+        # Upload combined stdout log
+        upload_artifact "${combined_stdout}" "test_artifacts/${test_id}/combined_stdout.log"
+
+        # Upload coredumps if any
+        for core_file in /tmp/coredumps/core.*; do
+            if [ -f "$core_file" ]; then
+                upload_artifact "$core_file" "test_artifacts/${test_id}/coredump"
+                break
+            fi
+        done
+
         sendLogFile
+        upload_artifact "${logFile}" "test_artifacts/${test_id}/full_output.log"
         postStatus "completed" "Ran all tests"
 
         sudo shutdown -h now
--- a/install/ci-vm/ci-linux/ci/runCI
+++ b/install/ci-vm/ci-linux/ci/runCI
@@ -128,8 +128,7 @@ if [ -e "${dstDir}/ccextractor" ]; then
         echo "=== CCExtractor Binary Version ===" >> "${logFile}"
         ./ccextractor --version >> "${logFile}" 2>&1
         echo "=== End Version Info ===" >> "${logFile}"
-
-
+        postStatus "testing" "Running tests"
         ccextractor_path="$(pwd)/ccextractor"
         combined_stdout="/tmp/combined_stdout.log"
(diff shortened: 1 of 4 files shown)


install/ci-vm/ci-linux/ci/runCI
```diff
@@ -124,8 +129,65 @@ if [ -e "${dstDir}/ccextractor" ]; then
...
+        chmod +x "${wrapper_path}"
+
         executeCommand cd ${suiteDstDir}
-        executeCommand ${tester} --debug --entries "${testFile}" --executable "ccextractor" --tempfolder "${tempFolder}" --timeout 600 --reportfolder "${reportFolder}" --resultfolder "${resultFolder}" --samplefolder "${sampleFolder}" --method Server --url "${reportURL}"
+        executeCommand ${tester} --debug --entries "${testFile}" --executable "${wrapper_path}" --tempfolder "${tempFolder}" --timeout 600 --reportfolder "${reportFolder}" --resultfolder "${resultFolder}" --samplefolder "${sampleFolder}" --method Server --url "${reportURL}"
+
+        # Upload AI artifacts to GCS
+        gcs_bucket=$(curl -s "http://metadata/computeMetadata/v1/instance/attributes/bucket" -H "Metadata-Flavor: Google")
+        test_id=$(curl -s "http://metadata/computeMetadata/v1/instance/attributes/testID" -H "Metadata-Flavor: Google")
+        token=$(curl -s "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google" | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])")
```
canihavesomecoffee: I'm not in favour of directly uploading to a bucket; I'd rather have it stored through the SP, where we can exercise more control and do extra checks. This direct upload would allow someone with malicious intentions to hijack a CCExtractor PR and start uploading other things.

mod_test/controllers.py
```diff
@@ -367,7 +367,6 @@ def generate_diff(test_id: int, regression_test_id: int, output_id: int, to_view
 
 
 @mod_test.route('/log-files/<test_id>')
-@login_required
```
canihavesomecoffee, line 370: A login should still be required.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+    )
+    return redirect(url)
+
+
+@mod_test.route('/<int:test_id>/binary', methods=['GET'])
+def download_binary(test_id):
+    """Download the ccextractor binary used in a test (linux or windows)."""
+    from run import storage_client_bucket
+    # Try linux name first, then windows
+    for name in ['ccextractor', 'ccextractor.exe']:
```
canihavesomecoffee: Based on the test ID, you can exactly know what platform it was for, and thus check the direct extension rather than "guessing".

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+        version="v4",
+        expiration=timedelta(minutes=config.get('GCS_SIGNED_URL_EXPIRY_LIMIT', 30)),
+        method="GET",
+        response_disposition=f'attachment; filename="{filename}"'
+    )
+    return redirect(url)
+
+
+@mod_test.route('/<int:test_id>/binary', methods=['GET'])
+def download_binary(test_id):
```
canihavesomecoffee, line 472: This needs a logged-in user.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+    # Try linux name first, then windows
+    for name in ['ccextractor', 'ccextractor.exe']:
+        blob_path = f'test_artifacts/{test_id}/{name}'
+        if storage_client_bucket.blob(blob_path).exists():
+            return _artifact_redirect(blob_path, filename=name)
+    abort(404)
+
+
+@mod_test.route('/<int:test_id>/coredump', methods=['GET'])
+def download_coredump(test_id):
```
canihavesomecoffee, line 483: This needs a logged-in user.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+    """Download the coredump from a test, if one was produced."""
+    return _artifact_redirect(
+        f'test_artifacts/{test_id}/coredump',
+        filename=f'coredump-{test_id}'
+    )
+
+
+@mod_test.route('/<int:test_id>/combined-stdout', methods=['GET'])
+def download_combined_stdout(test_id):
```
canihavesomecoffee, line 493: This needs a logged-in user.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+    """Download the combined stdout/stderr log from all test invocations."""
+    return _artifact_redirect(
+        f'test_artifacts/{test_id}/combined_stdout.log',
+        filename=f'combined_stdout-{test_id}.log'
+    )
+
+
+@mod_test.route('/<int:test_id>/regression/<int:regression_test_id>/<int:output_id>/output-got', methods=['GET'])
```
canihavesomecoffee, line 501: This needs a logged-in user.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+        abort(404)
+    ext = os.path.splitext(rf.regression_test_output.filename_correct)[1]
+    return _artifact_redirect(
+        f'TestResults/{rf.got}{ext}',
+        filename=f'output_got_{regression_test_id}_{output_id}{ext}'
+    )
+
+
+@mod_test.route('/<int:test_id>/regression/<int:regression_test_id>/<int:output_id>/output-expected', methods=['GET'])
+def download_output_expected(test_id, regression_test_id, output_id):
```
canihavesomecoffee, line 521: This needs a logged-in user.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+    )).first()
+    if rf is None:
+        abort(404)
+    ext = os.path.splitext(rf.regression_test_output.filename_correct)[1]
+    return _artifact_redirect(
+        f'TestResults/{rf.expected}{ext}',
+        filename=f'output_expected_{regression_test_id}_{output_id}{ext}'
+    )
+@mod_test.route('/<int:test_id>/sample/<int:sample_id>', methods=['GET'])
+def download_sample_ai(test_id, sample_id):
```
canihavesomecoffee, line 537: Even with AI workflow, there should be auth. We could perfectly issue a long(er) living access token for programmatic access. If we don't, a scraper might find this and still cause huge fees.

mod_test/controllers.py
```diff
@@ -442,3 +443,247 @@ def stop_test(test_id):
...
+    sample = Sample.query.filter(Sample.id == sample_id).first()
+    if sample is None:
+        abort(404)
+    return _artifact_redirect(
+        f'TestFiles/{sample.filename}',
+        filename=sample.original_name
+    )
+
+
+def _process_test_case(test_id, category_name, t_data):
```
canihavesomecoffee: Please fix the code smell here.

canihavesomecoffee (changes requested):

install/ci-vm/ci-linux/ci/runCI
pulk17: Sure, I understand the concern, will fix it.
Also can you please elaborate on what you mean by `stored through SP`?

PR #1070, opened 2026-03-20
[FIX] Guard .first() queries against None to prevent AttributeError

DhanushVarma-2:
Three queries called .first().value or .first().commit without checking
for None. On a fresh install or empty database, these crash with
AttributeError.

Fixed:
- mod_ci/controllers.py: GeneralData query for fetch_commit_{platform}.
  Downstream query for last_commit skipped when commit_hash is None
  to avoid WHERE commit IS NULL.
- mod_home/controllers.py: GeneralData query for last_commit on homepage.
  Template receives None which Jinja2 renders gracefully.
- mod_sample/controllers.py: GeneralData and CCExtractorVersion queries.
  Downstream Test queries skipped when values are None.


templates/sample/index.html
```diff
@@ -9,6 +9,8 @@
             <div class="column medium-8">
                 <h1>Sample information</h1>
                 <p>Please select one of the samples below to see more detailed information about it.</p>
+                {% set use_sample_original=true %}
+                {% set use_sample_original=true %}
```
canihavesomecoffee, line 13: Why where these added?

canihavesomecoffee (changes requested):

PR #1091, opened 2026-04-05
[FIX]: correct GitHub commit status when test output files are missing

x15sr71:

---

## Problem

`progress_type_request()` determined pass/fail for the GitHub commit status check using two raw SQL queries — one counting exit code mismatches, one counting non-null `TestResultFile.got` values. If the VM never wrote a `TestResultFile` row at all (missing output, failed mid-run), both counts return 0 and the platform posts a fake ✅ SUCCESS to GitHub even though no output was ever compared.

## Fix

Replace the dual-query block with `get_test_results(test)` — already used by the test detail page, `update_build_badge`, and `get_info_for_pr_comment`. It correctly treats a missing row as an error:
```python
if len(outputs) > 0:
    test_error = True  # output expected but never recorded → FAILURE
```

The result is also passed into `update_build_badge()` to avoid a redundant second call.

## Changes — `mod_ci/controllers.py` only, no schema change:

- Remove unused `from sqlalchemy.sql.functions import count`
- `update_build_badge(status, test)` → `update_build_badge(status, test, test_results=None)` — existing callers unaffected
- Replace dual-query block with `get_test_results` + `any(category['error'] for category in test_results)`

| Scenario | Before | After |
|---|---|---|
| Output correct | ✅ | ✅ |
| Output wrong | ❌ | ❌ |
| No output at all | ✅ wrong | ❌ correct |

Bot PR comments and test detail UI were already using `get_test_results` — both unaffected.


x15sr71:
Closing as #1119 supersedes this PR

PR #1094, opened 2026-04-07
[FIX] Fix negative duration values in CI test average time calculation

pulk17:

---

**Problem**
The platform was producing negative duration values for CI test runs, which corrupted the stored average time and caused negative estimated times to be displayed to users.
This stemmed from three separate bugs in update_final_status():

* Fragile timestamp parsing via group_concat — timestamps were concatenated into a string, split on ,, and manually parsed with a try/except for microsecond format. Since group_concat doesn't guarantee ordering, start/end could be swapped, producing a negative duration.
* Wrong test count formula — the incremental average used all_results / regression_test_count as a proxy for the number of finished tests per platform. This is completely unrelated to actual test history and produced a wrong weight in the running average.
* No negative guard on the final stored value — even if the computed duration was negative, it was written directly to the DB and later surfaced as a negative avg_minutes in the UI.

**Fix**

* Replaced group_concat + manual string parsing with func.min(timestamp) / func.max(timestamp) directly in SQL — unambiguous, ordering-independent, and format-independent.
* Replaced the broken test count with a proper COUNT(DISTINCT test_id) query filtered to completed/canceled progress entries for the current platform.
* Used the correct incremental mean formula: (old_avg × (n−1) + new_duration) / n.
* Added max(0, ...) guards on duration, last_duration, updated_average, and avg_minutes to prevent any negative value from propagating through to the DB or UI.

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -2526,29 +2526,17 @@ def progress_type_request(log, test, test_id, request) -> bool:
                     TestProgress.test_id.in_(platform_tests.select())
                 )
             )
-            in_progress_statuses = [TestStatus.preparation, TestStatus.completed, TestStatus.canceled]
-            finished_tests_progress = g.db.query(TestProgress).filter(
-                and_(
-                    TestProgress.test_id.in_(finished_tests.select()),
-                    TestProgress.status.in_(in_progress_statuses)
-                )
-            ).subquery()
             times = g.db.query(
-                finished_tests_progress.c.test_id,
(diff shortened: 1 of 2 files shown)


canihavesomecoffee (dismissed):

canihavesomecoffee:
Can you rebase and fix the merge conflicts?

pulk17:
done @canihavesomecoffee

canihavesomecoffee (approved):

PR #1117, opened 2026-06-04
[Feature] : Add REST API endpoints for access through AI  

pulk17:
Please prefix your pull request with one of the following: **[FEATURE]** **[FIX]** **[IMPROVEMENT]**.

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/sample-platform/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**My familiarity with the project is as follows (check one):**

- [ ] I have never used the project.
- [ ] I have used the project briefly.
- [ ] I have used the project extensively, but have not contributed previously.
- [x] I am an active contributor to the project.

---

# Feature: JSON REST API for the CCExtractor CI Platform (`mod_api`)

## Summary

This PR introduces a complete, spec-driven JSON REST API for the CCExtractor Sample Platform, mounted at `/api/v1`. It replaces the need for scrapers or direct database access by exposing every CI data point (runs, samples, results, diffs, logs, errors, artifacts, queue status) through authenticated, validated, and rate-limited endpoints.

The API is fully described by the `openapi-ci-api.yaml` specification and validated against it using Schemathesis property-based testing. It integrates cleanly with the existing data model (`mod_test`, `mod_regression`, `mod_sample`, `mod_auth`) without modifying any of those modules.

---

## Motivation

The existing platform serves server-rendered HTML pages and lacks a programmatic interface. This makes it impossible for:

- External dashboards and monitoring tools to query CI status
- Automated pipelines to trigger test runs or approve baselines
- Frontend rewrites (e.g., a React or Vue SPA) to consume structured data
- Third-party integrations to pull regression test results

This API solves all of the above by providing a standard REST interface with proper authentication, authorization, pagination, filtering, and error handling.

---

## Architecture

The module is organized into five layers, each with a single responsibility:

```
mod_api/
├── __init__.py          # Blueprint registration + import ordering
├── utils.py             # Pagination helpers, sort utilities
├── models/
│   └── api_token.py     # ApiToken SQLAlchemy model (argon2 hashed)
├── middleware/
│   ├── auth.py          # Bearer token auth + scope/role decorators
│   ├── error_handler.py # Centralized JSON error responses
[...]

PR #1130, opened 2026-06-24
[Feature] : API ENDPOINTS PR 1: Foundation and scaffolding

pulk17:
**[Feature]**

---

## Overview

First PR of a 7-part stack that adds a token-authenticated JSON REST API (`/api/v1`) to the sample platform, giving CI consumers and future tooling programmatic access to runs, samples, results, and logs — everything that today requires scraping the HTML views.

**Stack: #1130 → #1131 → #1132 → #1133 → #1134 → #1135 → #1141. Merge bottom-up.** Each PR contains the ones below it, so the diff shrinks automatically as parents merge.

This PR deliberately contains **no endpoints**. It lays down the infrastructure the six route PRs build on, so their reviews can focus on endpoint logic.

## What's in it

**Blueprint + middleware** (`mod_api/`)
- `mod_api` blueprint registered at `/api/v1`. Anything under that prefix returns structured JSON errors (`{code, message, details}`) — including routing-level 404/405s, which are converted app-wide via an `after_app_request` hook so no HTML error page ever leaks into an API response.
- **Bearer auth**: `Authorization: Bearer spci_...` (scheme matched case-insensitively per RFC 7235). Tokens are 256-bit random secrets; the server stores only a SHA-256 hash plus an indexed 16-char prefix for lookup. Verification fetches candidates by prefix and compares hashes with `hmac.compare_digest` (constant-time). Scope checks (`@require_scope`) and role checks (`@require_roles`) are separate decorators, so each endpoint's requirements are visible at its definition.
- **Rate limiting**: per-client sliding window with standard `X-RateLimit-*` response headers. In-memory by design — single-process deployment; the store interface is small enough to swap for Redis if that ever changes.
- **Security headers** on every API response; hook registration order (and Flask's reverse execution order for `after_request`) is documented in `mod_api/__init__.py`.
- **Validation helpers**: offset/cursor pagination, path-id, date-range, and sort-parameter validators used consistently by all later PRs.

**Model + migration**
- `ApiToken` model: name (unique per user, kept after revocation for audit), hash, prefix, JSON scopes, expiry, revocation timestamps.
- Alembic migration `d4f8e2a1b3c7`: new `api_token` table + nullable `user.github_login` column. Purely additive — no existing table is modified, so it's safe for the deployment pipeline to auto-apply, and rolling back the code without rolling back the schema is harmless.

**Status derivation service** (`mod_api/services/status.py`)
The single source of truth mapping raw `TestProgress`/`TestResult`/`TestResultFile` rows to normalized statuses (runs: `queued/running/pass/fail/canceled/error/incomplete`; samples: `pass/fail/skipped/missing_output/running/not_started`). Route handlers are forbidden from inlining their own derivation — this is what keeps `/summary`, `/samples`, and `/errors` from ever disagreeing. Documented data-model traps it encodes: `TestResultFile.got = NULL` means *match*, not missing; the `(-1, 'error')` dummy row means the test produced no output; `test.failed` only reflects cancellation. A run marked completed that produced zero result rows reports `error`, never `pass`.

**Test scaffolding**
`tests/api/base.py` provides `ApiTestCase`, which stubs the deliberately-slow `sha512_crypt` password hashing for API tests (they create users/tokens constantly). This keeps the API suite in the seconds range instead of minutes; any test needing real hashing can stop the patchers locally.

## Testing

42 tests: token model (generation, hashing, prefix extraction, expiry), status derivation (including the completed-without-results → `error` case), validation helpers, and utils. Full lint battery (isort, pycodestyle, pydocstyle, dodgy, mypy) is green.

### DIFF
--- a/.pycodestylerc
+++ b/.pycodestylerc
@@ -1,5 +1,5 @@
 [pycodestyle]
 count = True
 max-line-length = 120
-exclude=test_diff.py,migrations,venv*,parse.py,config.py
+exclude=test_diff.py,migrations,venv*,.venv*,parse.py,config.py
 ignore = E701
--- a/mod_api/models/api_token.py
+++ b/mod_api/models/api_token.py
@@ -1,0 +1,175 @@
+"""
+ApiToken model: server-side storage for scoped API tokens.
+
+Tokens are opaque strings prefixed with 'spci_'. Only the SHA-256 hash
+is persisted; the plaintext is returned exactly once at creation time.
+A fast hash with a constant-time compare is sufficient here because the
+tokens are 256-bit random secrets — a slow password KDF (argon2/bcrypt)
+buys nothing against brute force on high-entropy values.
+"""
+
+import hashlib
+import hmac
+import json
+import secrets
+from datetime import datetime, timedelta, timezone
+from typing import List
+
+from sqlalchemy import (Column, DateTime, ForeignKey, Integer, String, Text,
+                        UniqueConstraint)
+from sqlalchemy.orm import relationship, validates
+
+from database import Base
+
+
+class Scope:
+    """The scopes a token can carry.
+
+    Reference these constants instead of writing the strings inline, so a
+    rename stays a single edit. Plain strings rather than a DeclEnum
+    because scopes cross the wire: they arrive in request bodies, are
+    stored as a JSON array, and are compared against what the client sent.
+    """
+
+    RUNS_READ = 'runs:read'
+    RUNS_WRITE = 'runs:write'
+    RESULTS_READ = 'results:read'
+    BASELINES_WRITE = 'baselines:write'
+    SYSTEM_READ = 'system:read'
+    TOKENS_MANAGE = 'tokens:manage'
+
+
+VALID_SCOPES = frozenset([
+    Scope.RUNS_READ,
+    Scope.RUNS_WRITE,
+    Scope.RESULTS_READ,
+    Scope.BASELINES_WRITE,
+    Scope.SYSTEM_READ,
+    Scope.TOKENS_MANAGE,
+])
+
+DEFAULT_SCOPES = [Scope.RUNS_READ, Scope.RESULTS_READ]
+
+TOKEN_PREFIX = 'spci_'
+TOKEN_BYTE_LENGTH = 32
+
+
+class ApiToken(Base):
+    """Scoped API token bound to a user account."""
+
+    __tablename__ = 'api_token'
+    __table_args__ = (
+        UniqueConstraint('user_id', 'token_name', name='uq_user_token_name'),
+        {'mysql_engine': 'InnoDB'},
+    )
+
+    id = Column(Integer, primary_key=True)
+    user_id = Column(
+        Integer,
+        ForeignKey('user.id', onupdate='CASCADE', ondelete='CASCADE'),
+        nullable=False,
+    )
+    user = relationship('User', uselist=False)
+    token_name = Column(String(50), nullable=False)
+    token_hash = Column(String(255), nullable=False)
+    token_prefix = Column(String(16), nullable=False, index=True)
+    scopes_json = Column(Text(), nullable=False)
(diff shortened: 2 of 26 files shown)


cfsmp3 (changes requested):
Note the two high complexity errors (validation.py and status.py), also see Claude review below.

Reading the rview, clearly H1 must be addressed before we can merge or things will break.

The rest could be addressed (possibly are) later on in the stack.

But let's make sure we can merge and test each PR in order but not at the same time, i.e. merging this diff shouldn't break the system, even if by itself it doesn't do anything useful (since it's scaffolding).

Claude review follows:

HIGH (blocker):
- H1 — ApiToken model added with NO migration. The api_token table won't exist on real MySQL. Tests pass only because tests/base.py does create_all from models → masks the gap. PR2's auth endpoints break at runtime. #1117 had this migration (d4f8e2a1b3c7); dropped in the split. Must add it, chained off master head c8f3a2b1d4e5.

MEDIUM (carryover, code unchanged):
- ~770 lines of security middleware merge untested — and with no routes yet, the before_request hooks never even fire in this PR. Deferred to PR2/3. Risk noted.
- Rate limiter unbounded memory (no hard cap, eviction only every 100 req).
- Auth timing oracle (no-candidate path skips argon2 verify → leaks whether a prefix exists).
- _get_client_ip comment wrong (ProxyFix means remote_addr is from XFF).

LOW/NIT: run-level missing→fail path untested; stale is_dummy_row "DEPLOYMENT PREREQUISITE" docstring; N+1 footgun in the batch_get_run_data wrappers; pytest conftest.py inert under nose2; generic 429 handler hardcodes wrong limits.

cfsmp3:
New / still open:
- 🔴 N1 (Medium) — the timing-attack fix doesn't work. They added a dummy argon2 verify on the no-candidate path, but the hardcoded hash is malformed — I measured it: verify fails-fast in 0.0 ms vs ~43 ms for a real verify. So the timing oracle is still open, now hidden behind code that looks like a fix. Cheap fix: _DUMMY_HASH = PasswordHasher().hash('dummy') at import. (Practical risk is low — prefixes are 65-bit — but don't ship dead security code.)
- 🟡 N2 (Medium) — migration/model drift: the migration adds user.github_login but mod_auth/models.py doesn't declare it → tests (create_all) won't have the column, and a future flask db migrate would propose dropping it. Add the field to the User model here, or defer that migration line to #1132 where it's used.
- 🟡 M2 (rate-limit memory cap) still open.

.pycodestylerc
```diff
@@ -1,5 +1,5 @@
 [pycodestyle]
 count = True
 max-line-length = 120
-exclude=test_diff.py,migrations,venv*,parse.py,config.py
-ignore = E701
+exclude=test_diff.py,migrations,venv*,.venv*,parse.py,config.py
+ignore = E701,W503
```
canihavesomecoffee: Any reason why you want to ignore this warning?

canihavesomecoffee (approved):
If the pending comments are resolved it's good to go :)

mod_api/models/api_token.py
```diff
@@ -0,0 +1,157 @@
...
+from database import Base
+
+VALID_SCOPES = frozenset([
+    'runs:read',
+    'runs:write',
+    'results:read',
+    'baselines:write',
+    'system:read',
+    'tokens:manage',
+])
```
canihavesomecoffee: These should be moved to either string constants or (preferred) an enum-like object, so you can reference them everywhere with a single defined string (similar to a Role).

canihavesomecoffee (changes requested):

.pycodestylerc
pulk17: No particular reason, forgot to cleanup after some local testing

mod_api/models/api_token.py
pulk17: Done

canihavesomecoffee (approved):

PR #1131, opened 2026-06-24
[Feature] : API ENDPOINTS PR 2: Auth

pulk17:
[FEATURE]

---

## Overview

Second PR of the CI API stack (#1130 → #1141, merge bottom-up). Builds on the auth middleware from #1130 and adds the token lifecycle — the only part of the API usable with just an email and password; everything else requires a token minted here.

## Endpoints

| Endpoint | Auth | Purpose |
|---|---|---|
| `POST /auth/tokens` | email + password | Mint a scoped bearer token |
| `GET /auth/tokens` | `tokens:manage` + admin role | List tokens (own or `?all=true`) |
| `DELETE /auth/tokens/current` | any valid token | Self-revocation |
| `DELETE /auth/tokens/{id}` | `tokens:manage` + admin role | Revoke by id |

## Design decisions

- **The plaintext token is returned exactly once**, in the creation response, and never stored or shown again. List responses contain metadata only — no token material, not even the prefix.
- **Scopes are validated against the requester's role at mint time**: a contributor cannot mint `baselines:write` even by asking for it (403 with the missing scopes in `details`). Default expiry 7 days, max 30.
- **Self-revocation is deliberately scope-free**: possession of a valid token is sufficient authority to destroy that token. Requiring `tokens:manage` here would force every automation token to carry an admin-level scope just to be able to clean up after itself. Revoke-by-id, which acts on *other* tokens, does require it.
- **Token names stay reserved after revocation** — revoked rows are kept for audit, and the unique constraint spans them. The 409-style conflict message states this explicitly instead of leaving users confused about a name they "deleted". Uniqueness is also enforced at the DB level, and the `IntegrityError` path is handled (and tested) for races between the pre-check and the insert.

## Also in this PR

- Integration tests for the error-handler and rate-limit middleware (they need a live endpoint to exercise; `POST /auth/tokens` is the first).
- One-line `tests/test_ci` fix: `test_webhook_prerelease` now mocks `requests.get` like its sibling webhook tests, removing a network dependency.

## Testing

23 new tests: happy paths, invalid credentials, role/scope rejection, duplicate names, the IntegrityError race, revocation flows (revoked token immediately 401s), and rate-limit/error-handler behavior. 65 total at this layer.


cfsmp3 (changes requested):
Missing tests:

- Rate limiter still untested (tests bypass it via TESTING + _rate_limit_store.clear()). The 5/15min create limit is unverified.
- H4 fix (500→JSON) still unverified — no test forces a real route 500. 

Bug:
- Non-admins can still request tokens:manage scope (harmless given role gates, but loose).

mod_api/routes/auth.py
```diff
@@ -0,0 +1,213 @@
...
+    Non-admins can only revoke their own tokens. Admins can revoke anyone's.
+    Already-revoked tokens are silently accepted (idempotent).
+
+    Deliberately requires no extra scope: scopes gate data access, while
+    revocation is self-service credential hygiene. Any valid token may
+    revoke tokens belonging to its own user — plain users cannot obtain
+    tokens:manage (see create_token), yet must be able to clean up their
+    own credentials.
+    """
+    is_admin = g.api_user.role.value == 'admin'
```
canihavesomecoffee: Why not use the `is_admin` property of the user?

mod_api/routes/auth.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Check role limitations
+    # Note: Plain 'user' role deliberately cannot request tokens:manage. They
+    # can create tokens with runs:write but cannot list them. They must revoke
+    # either the current token or by ID.
+    allowed_scopes = {
+        'runs:read', 'runs:write', 'results:read',
+        'system:read'
+    }
+    if user.role.value == 'admin':
```
canihavesomecoffee: Same here, why not use `user.is_admin`?

mod_api/routes/auth.py
```diff
@@ -0,0 +1,213 @@
...
+            http_status=401,
+        )
+    token.revoke()
+    g.db.add(token)
+    g.db.commit()
+    return '', 204
+
+
+@mod_api.route('/auth/tokens', methods=['GET'])
+@require_roles(['admin'])
```
canihavesomecoffee: I'd rather see something similar to what we have, with a Role passed to the decorator rather than a string. If we ever change the names of the roles (for whatever reason), this might cause failures without anyone noticing.

mod_api/routes/auth.py
```diff
@@ -0,0 +1,213 @@
...
+@require_scope('tokens:manage')
```
canihavesomecoffee: I'd rather see a reference to a predefined scope here (same argumentation as for the role)

canihavesomecoffee (changes requested):

canihavesomecoffee (approved):

PR #1132, opened 2026-06-24
[Feature] : API ENDPOINTS PR 3 : System Status and Run Management Endpoints

pulk17:
[FEATURE]

---

## Overview

Third PR of the CI API stack (#1130 → #1141, merge bottom-up). The core read/write surface for CI runs: monitoring endpoints, run listing and detail, and the ability to trigger and cancel runs via the API.

## Endpoints

**System**
- `GET /system/health` — unauthenticated, for monitors and load balancers. Reports overall `ok/degraded/down` plus per-dependency status (database, local storage, GCS). 503 only when down.
- `GET /system/queue` — queued and running work with `queue_depth`/`running_count` meta and `?platform`/`?status` filters. Queue `position` is only populated under `?status=queued`, where it's well-defined.

**Runs**
- `GET /runs` — filterable by status, platform, branch, repository, and creation date range
- `POST /runs` — trigger a run for a commit (`runs:write`)
- `GET /runs/{id}`, `POST /runs/{id}/cancel`, `GET /runs/{id}/summary`, `GET /runs/{id}/config`, `GET /runs/{id}/progress`

## Implementation notes for reviewers

- **`?status=queued` composes with date filters.** Queued runs have no progress rows yet, so the date subquery LEFT JOINs progress and treats "no first-progress timestamp" as matching. An inner join here would make `?status=queued&created_after=...` silently return nothing.
- **`POST /runs` permission model**: the requesting user's GitHub identity must match the target repository (fork owner, or the main repo for maintainers — compared case-insensitively, since GitHub owner names are). The OAuth callback in `mod_auth` now records `github_login` to support this; this PR is its first consumer. Run creation also verifies a CI build artifact exists for the commit before enqueueing, so the API can't create runs that would instantly fail.
- **Summary/config count only active regression tests**, and customized runs count their selected subset — matching what the worker will actually execute, not the whole catalog.
- **Error semantics**: a completed run with zero `TestResult` rows is reported as `error` (worker finished without reporting), never `pass`.
- All timestamps are UTC with the same `Z`-suffixed format used across every endpoint in the stack.
- `error_service` (structured error derivation, used by run summaries here and expanded in #1135) and `storage` (GCS signed-URL resolution with local-file fallback) services land here with their consumers.

The auth middleware integration tests also land in this PR — they need a stable protected endpoint to probe, and `GET /system/queue` is it (401 paths: missing/malformed header, bad prefix, wrong hash, revoked, expired; 403 scope paths; prefix-collision resolution; `g.api_user` population).

## Testing

61 new tests; 126 total at this layer. Query composition (status × date), permission rejections, cancellation flow, summary/config counting, health degradation paths, and the full middleware-auth matrix.

### DIFF
--- a/mod_api/__init__.py
+++ b/mod_api/__init__.py
@@ -36,3 +36,5 @@ mod_api.after_app_request(error_handler.convert_api_errors_to_json)
 # Route modules register themselves against the blueprint; the rest of
 # the stack adds one module per PR.
 from mod_api.routes import auth as auth_routes  # noqa: E402, F401
+from mod_api.routes import runs as runs_routes  # noqa: E402, F401
+from mod_api.routes import system as system_routes  # noqa: E402, F401
--- a/mod_api/routes/runs.py
+++ b/mod_api/routes/runs.py
@@ -1,0 +1,677 @@
+"""
+Test run routes.
+
+GET    /runs                   List runs (filtered, paginated, sorted)
+POST   /runs                   Trigger a new run
+GET    /runs/{id}              Single run details
+GET    /runs/{id}/summary      Pass/fail/skip counts
+GET    /runs/{id}/progress     Progress event timeline
+GET    /runs/{id}/config       Run configuration and test matrix
+POST   /runs/{id}/cancel       Cancel a queued or running test
+"""
+
+from collections import defaultdict
+
+from flask import g, request
+from sqlalchemy import func, or_
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import joinedload
+
+from mod_api import mod_api
+from mod_api.middleware.auth import require_roles, require_scope
+from mod_api.middleware.error_handler import make_error_response
+from mod_api.middleware.validation import (validate_body, validate_date_range,
+                                           validate_offset_pagination,
+                                           validate_path_id, validate_sort)
+from mod_api.models.api_token import Scope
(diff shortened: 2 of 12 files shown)


cfsmp3 (changes requested):
HIGH:
- H2 — the C1 fix is only half-wired. The expected_outputs fix went into batch_get_run_data (run status), but get_run_summary and error_service still call the legacy is_dummy_row path (never matches). So for a run with missing output: derive_run_status says fail (right), but /summary counts it as pass with missing_output_count: 0, and /errors omits it. The endpoints contradict each other, and the headline bug resurfaces in the two endpoints users actually read. No test catches it.
- H3 — arbitrary-repo CI trigger (carryover #1117 H2). POST /runs: is_staff short-circuits ownership, repository is only regex-checked → a contributor can make the build VM clone+build any GitHub repo = arbitrary-code exec on CI infra. Gated by contributor role, but a real privilege expansion.
- H1 — migration still absent (3 PRs deep; all these routes are auth-gated so they need the token table too).

MEDIUM: cancel race (no row lock); status-filter 1000-cap drops rows + inconsistent total; summary total_samples never reconciles; _get_output_artifacts N+1 + paginates after full build; GCS signed URLs without blob.exists.

LOW/NIT: RunSummarySchema unused (raw dict, no contract); step:None dead field; queue hand-builds pagination; generic-exception→JSON path still untested.

canihavesomecoffee (approved):

PR #1142, opened 2026-06-30
feat: smart (semantic) diff for subtitle regression outputs

gaurav02081:
## What this is

A **semantic diff** for regression-test outputs. Instead of a raw line diff, it
classifies *how* two subtitle outputs differ, so a person — or an AI agent —
gets an actionable verdict instead of a wall of changed lines.

Example: instead of "53 lines differ", it says **"All 13 cues match but are
500 ms late."**

## Classification kinds

`identical`, `timing_shift` (constant offset), `timing_drift` (growing offset),
`text_change`, `formatting_change` (tags/entities only), `whitespace_change`
(CEA-608 padding only), `encoding_change` (non-ASCII/accents only), `split_cues`,
`merged_cues`, `missing_cues`, `extra_cues`, `unsupported` (no cues parsed), or
`mixed`. Each result also carries a per-cue `changes` list (which cue, what kind,
expected/actual snippet, timing offset).

## Grounded in CCExtractor's own behaviour

- Mirrors `tests/extract_expected.py`'s normalization (strip styling tags,
  unescape entities, trim CEA-608 trailing padding) so cosmetic differences are
  separated from real text changes.
- Handles the `-latin1` charset case (accent folding → `encoding_change`).
- Verified on **real** CCExtractor output (CEA-608 English + DVB Spanish with
  `<font>` tags and accents), vendored as golden fixtures.

## Where it lives

- `mod_test/smartdiff/` — a Flask-decoupled core (SRT + WebVTT parsers, a format
  dispatcher, normalization, and the classifier). Fully unit-tested.
- `TestResultFile.generate_smart_diff()` + a JSON endpoint
  `GET /diff/<test>/<regression>/<output>/smart` — reusable by the web UI, a CLI,
  and agents.
- A **"Smart"** option next to the existing "Fail" diff link on the result page
  (additive; the raw diff is unchanged).

## Notes

- **Independent** of the REST API work — this is standalone in the platform.
- Tests: parsers, classifier, normalization, real golden fixtures, robustness
  (malformed/garbage input), and the model glue.
- Follow-up (not in this PR): a CLI `sp run diff --smart` consuming the same
  endpoint.

### DIFF
--- a/mod_test/controllers.py
+++ b/mod_test/controllers.py
@@ -375,6 +375,44 @@ def generate_diff(test_id: int, regression_test_id: int, output_id: int, to_view
     abort(404)
 
 
+@mod_test.route('/diff/<test_id>/<regression_test_id>/<output_id>/smart')
+def smart_diff_view(test_id: int, regression_test_id: int, output_id: int):
+    """
+    Return a semantic (smart) diff classification for an output as JSON.
+
+    Unlike the line diff, this reports *how* the output differs (timing shift,
+    cosmetic padding/formatting/encoding, text change, missing/extra cues), so a
+    person or an agent gets an actionable answer instead of a wall of lines.
+
+    :param test_id: id of the test
+    :type test_id: int
+    :param regression_test_id: id of the regression test
+    :type regression_test_id: int
+    :param output_id: id of the generated output
+    :type output_id: int
+    :return: JSON classification of the difference.
+    :rtype: flask.Response
+    """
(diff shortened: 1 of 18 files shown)


gaurav02081:
@canihavesomecoffee   Hi could you take a look at this PR? It just hasn't had eyes on it yet. Also, the SP CLI is live. Once this feature is merged I will wire it into that as well.

PR #1160, opened 2026-07-30
[Feature] : API ENDPOINTS PR 8: NG endpoints

pulk17:
[FEATURE]

---

## PR 8: Administration and Web Console Endpoints

> **Stacked PR.** This branches off PR 7 (#1141), which branches off PR 6 (#1135).
> Until those merge the diff shown here includes their commits — **only the final
> commit, `PR 8: Administration and Web Console Endpoints`, belongs to this PR.**
> It collapses to that one commit automatically as the earlier PRs land.

This is the last endpoint PR of the series. It closes the remaining gaps between
the JSON API and what the classic site can do, so a client can drive the platform
without falling back to the Jinja pages or to hand-written SQL.

Twenty operations, and only one new route module. `/auth/me` and `/users` go into
`routes/auth.py`, which already owns identity and token administration; platform
configuration goes into `routes/system.py`, which already owns the `/system`
prefix. Only the regression suite gets its own module, because `routes/samples.py`
is already 700 lines.

### What this adds

**Identity**

| Method | Endpoint | Access |
|---|---|---|
| `GET` | `/auth/me` | any valid token |

A token carries no role, so a client cannot tell what the caller may do. This
returns the account and its role, and echoes the token's granted scopes so a
client can also hide what this particular token could never call.

**Regression suite** — replaces editing the tables by hand

| Method | Endpoint | Access |
|---|---|---|
| `GET` | `/regression-tests/{id}` | `runs:read` |
| `POST` | `/regression-tests` | admin/contributor + `runs:write` |
| `PATCH` | `/regression-tests/{id}` | admin/contributor + `runs:write` |
| `DELETE` | `/regression-tests/{id}` | admin/contributor + `runs:write` |
| `GET` | `/categories` | `runs:read` |
| `POST` | `/categories` | admin/contributor + `runs:write` |
| `PATCH` | `/categories/{id}` | admin/contributor + `runs:write` |
| `DELETE` | `/categories/{id}` | admin/contributor + `runs:write` |

**Samples**

| Method | Endpoint | Access |
|---|---|---|
| `GET` | `/samples/{id}/details` | `runs:read` |

Upload record, extra files and the parsed MediaInfo tree — the detail view the
summary endpoint does not cover.

**Users**

| Method | Endpoint | Access |
|---|---|---|
| `GET` | `/users` | admin + `tokens:manage` |
| `PATCH` | `/users/{id}` | admin + `tokens:manage` |

**Platform configuration** — the equivalents of the maintenance and blocked-user
admin pages

| Method | Endpoint | Access |
|---|---|---|
| `GET` | `/system/maintenance` | `system:read` |
| `PATCH` | `/system/maintenance/{platform}` | admin + `system:write` |
| `GET` | `/system/blocked-users` | `system:read` |
| `POST` | `/system/blocked-users` | admin + `system:write` |
| `DELETE` | `/system/blocked-users/{user_id}` | admin + `system:write` |
| `GET` | `/system/forbidden-extensions` | `system:read` |
| `POST` | `/system/forbidden-extensions` | admin + `system:write` |
| `DELETE` | `/system/forbidden-extensions/{extension}` | admin + `system:write` |

### One new scope

`system:write` is added for platform configuration, and is admin-only at the
point a token is minted rather than only at the route. It is kept apart from
`system:read` so a monitoring token can watch the platform without being able to
pause CI. The scope-count cap on token creation is now derived from the scope
list itself, so adding a scope can't silently make "ask for everything I'm
allowed" fail validation.

Everything else reuses the existing scopes: suite editing is `runs:write`, and
user administration is `tokens:manage` rather than a new scope for one surface.
Happy to split that out if you would prefer a dedicated one.

### Behaviour worth reviewing

- **Deletes only remove what nothing references.** A regression test that has
  already run is refused with 409 and the number of results referencing it —
  removing it would erase evidence of past regressions. Retiring such a test is
  `PATCH {"active": false}`, which is what you actually want. A category still
  attached to tests is refused the same way, because dropping it would change
  which tests a suite selection picks up; detach them by PATCHing each test's
  categories. There is deliberately no force flag: nothing in this API can
  destroy run history.
- **Baselines, variants and category links are cleared before a delete.** Those
  foreign keys are `RESTRICT` with no ORM cascade, so the classic
  `mod_regression` delete raises as soon as a test has any output row.
- **New regression tests are created inactive** unless asked otherwise, so a test
  cannot join a CI suite before someone has seen what it actually produces.
- **Unknown category names reject the whole request** with the offending names in
  `details.fields.categories`, rather than silently dropping one.
- **`PATCH` writes only the fields present**, so two clients editing different
  fields cannot clobber each other. An empty body is a 400.
- **Admins cannot change their own role** — demoting the last admin here would
  leave nobody able to undo it.
- **Blocked users are keyed on the numeric GitHub id**, not the login: logins can
  be changed and reused, which would silently unblock somebody.
- **Media info is best-effort** — a sample whose XML is missing or unparseable
  reports `null` instead of failing. Unlike the classic page this never
  regenerates the XML, because a `GET` should not write to the sample repository.

### Response shapes

`/users`, `/categories`, `/system/blocked-users` and `/system/forbidden-extensions`
are offset-paginated like every other collection in this API.

`/system/maintenance` is the one exception and is keyed on `platforms` rather than
`data`: the list is derived from the `TestPlatform` enum, not a growable table, so
a `next_offset` would be meaningless. That matches how `/system/health` returns
`dependencies`.

### Contract

`openapi-ci-api.yaml` grows from 27 paths / 29 operations to 40 / 49, with a
`Users` tag, a shared `Conflict` response, five new parameters and fourteen new
schemas. Request bodies use `additionalProperties: false` with length and range
limits, matching the conventions PR 6 established.

Spec and implementation were cross-checked against the app's live `url_map`:
every documented operation is implemented and every implemented operation is
documented, in both directions, with no dangling `$ref`s.

`scripts/verify_schemathesis.py` needed two updates to stay usable. Its admin
fixture hardcodes a scope list, so without `system:write` every platform-config
write would have 403'd through the whole contract run. And its broad fuzz pass
loads the full spec while excluding only the destructive auth routes — the new
administration deletes and the maintenance switch are excluded for the same
reason, since fuzzing `PATCH /system/maintenance/{platform}` would pause CI on
whatever environment the harness runs against.

### Testing

67 new tests. Beyond the success paths they cover 400/401/403/404/409 for each
endpoint, role-versus-scope separation (a plain user holding `runs:write` is
still refused), the self-role-change refusal, that a delete is refused while
history exists, and that `system:write` cannot be minted by a non-admin.

### DIFF
--- a/mod_api/routes/regression_tests.py
+++ b/mod_api/routes/regression_tests.py
@@ -1,0 +1,379 @@
+"""
+Regression test detail/write endpoints and the categories that group them.
+
+GET    /regression-tests/{id}   Full detail, including baselines and variants
+POST   /regression-tests        Create a test (inactive unless asked otherwise)
+PATCH  /regression-tests/{id}   Partially update a test
+DELETE /regression-tests/{id}   Delete a test that has never run
+GET    /categories              List categories with their test counts
+POST   /categories              Create a category
+PATCH  /categories/{id}         Rename or re-describe a category
+DELETE /categories/{id}         Delete a category no test references
+
+Editing the suite previously meant hand-written SQL against the production
+database. The regression test list view stays in routes/samples.py.
+"""
+
+from flask import g
+from sqlalchemy import func
+from sqlalchemy.exc import IntegrityError
+
+from mod_api import mod_api
+from mod_api.middleware.auth import require_roles, require_scope
+from mod_api.middleware.error_handler import make_error_response
+from mod_api.middleware.validation import (validate_body,
+                                           validate_offset_pagination,
+                                           validate_path_id)
+from mod_api.models.api_token import Scope
+from mod_api.routes.samples import serialize_rt
+from mod_api.schemas.regression_tests import (CategoryCreateSchema,
+                                              CategoryUpdateSchema,
+                                              RegressionTestCreateSchema,
+                                              RegressionTestUpdateSchema)
+from mod_api.utils import paginated_response, single_response
+from mod_auth.models import Role
+from mod_regression.models import (Category, InputType, OutputType,
+                                   RegressionTest, RegressionTestOutput,
+                                   RegressionTestOutputFiles,
+                                   regressionTestLinkTable)
+from mod_sample.models import Sample
+from mod_test.models import TestResult
+
+
+def _resolve_categories(names):
+    """
+    Look up categories by name.
+
+    Returns the matching rows plus the names that matched nothing, so the
+    caller can reject the whole request rather than silently dropping a
+    category the client believed it had set.
+    """
+    rows = Category.query.filter(Category.name.in_(names)).all()
+    found = {category.name for category in rows}
+    return rows, [name for name in names if name not in found]
+
+
+def _unknown_categories_response(unknown):
+    """Build the 400 returned when a request names categories that don't exist."""
+    return make_error_response(
+        'validation_error',
+        f"Unknown categories: {', '.join(unknown)}",
+        details={'fields': {'categories': unknown}},
+        http_status=400,
+    )
+
+
+@mod_api.route('/regression-tests/<regression_test_id>', methods=['GET'])
+@require_scope(Scope.RUNS_READ)
+@validate_path_id('regression_test_id')
+def get_regression_test(regression_test_id):
+    """
+    Return one regression test with its baselines.
+
+    The list endpoint omits outputs because they multiply the payload for
+    every row; a detail view needs them to show what "expected" means,
+    including the alternative hashes accepted as variants.
+    """
+    test = RegressionTest.query.filter(
+        RegressionTest.id == regression_test_id).first()
+    if test is None:
+        return make_error_response(
+            'not_found', f'Regression test {regression_test_id} not found.',
+            http_status=404)
+
+    data = serialize_rt(test)
+    data['outputs'] = [
+        {
+            'id': output.id,
+            'correct': output.correct,
+            'correct_extension': output.correct_extension,
+            'expected_filename': output.expected_filename,
+            'ignore': output.ignore,
+            'variants': [f.file_hashes for f in output.multiple_files],
+        }
+        for output in test.output_files
+    ]
+    return single_response(data)
+
+
+@mod_api.route('/regression-tests', methods=['POST'])
+@require_roles([Role.contributor, Role.admin])
+@require_scope(Scope.RUNS_WRITE)
+@validate_body(RegressionTestCreateSchema)
+def create_regression_test(validated_data=None):
+    """
+    Create a regression test.
+
+    Created inactive unless the body says otherwise, so it cannot join a CI
+    suite before someone has seen the output it actually produces.
+    """
+    data = validated_data
+
+    sample = Sample.query.filter(Sample.id == data['sample_id']).first()
+    if sample is None:
+        return make_error_response(
+            'not_found', f"Sample {data['sample_id']} not found.",
+            http_status=404)
+
+    categories, unknown = _resolve_categories(data['categories'])
+    if unknown:
+        return _unknown_categories_response(unknown)
+
+    test = RegressionTest(
+        sample_id=sample.id,
+        command=data['command'],
+        input_type=InputType.from_string(data['input_type']),
+        output_type=OutputType.from_string(data['output_type']),
+        # Categories live in a link table; the constructor argument is a
+        # leftover from the single-category era and maps to no column.
+        category_id=None,
+        expected_rc=data['expected_rc'],
+        active=data['active'],
+        description=data['description'],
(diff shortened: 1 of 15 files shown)


mod_api/routes/system.py
```diff
@@ -345,3 +366,171 @@ def list_artifacts(run_id, limit=50, offset=0):
...
+        row.disabled = disabled
+    g.db.commit()
+
+    g.log.info(f'maintenance for {target.value} set to disabled={disabled} '
+               f'via API by user {g.api_user.id}')
+    return single_response(_maintenance_entry(target, row))
+
+
+@mod_api.route('/system/blocked-users', methods=['GET'])
+@require_scope(Scope.SYSTEM_READ)
```
canihavesomecoffee, line 439: A regular user should not be able to see what users are blocked.

mod_api/routes/regression_tests.py
```diff
@@ -0,0 +1,349 @@
...
+        for output in test.output_files
+    ]
+    return single_response(data)
+
+
+@mod_api.route('/regression-tests', methods=['POST'])
+@require_roles(_WRITE_ROLES)
+@require_scope(Scope.RUNS_WRITE)
+@validate_body(RegressionTestCreateSchema)
+def create_regression_test(validated_data=None):
```
canihavesomecoffee, line 103: This should be limited to contributors, testers or admins, not any user.

mod_api/routes/regression_tests.py
```diff
@@ -0,0 +1,349 @@
...
+    g.log.info(f'regression test {test.id} created via API by {g.api_user.id}')
+    return single_response(serialize_rt(test), http_status=201)
+
+
+@mod_api.route('/regression-tests/<regression_test_id>', methods=['PATCH'])
+@require_roles(_WRITE_ROLES)
+@require_scope(Scope.RUNS_WRITE)
+@validate_path_id('regression_test_id')
+@validate_body(RegressionTestUpdateSchema)
+def update_regression_test(regression_test_id, validated_data=None):
```
canihavesomecoffee, line 147: Same.

mod_api/routes/regression_tests.py
```diff
@@ -0,0 +1,349 @@
...
+    g.log.info(f'regression test {test.id} updated via API by '
+               f'{g.api_user.id}: {sorted(data.keys())}')
+    return single_response(serialize_rt(test))
+
+
+@mod_api.route('/regression-tests/<regression_test_id>', methods=['DELETE'])
+@require_roles(_WRITE_ROLES)
+@require_scope(Scope.RUNS_WRITE)
+@validate_path_id('regression_test_id')
+def delete_regression_test(regression_test_id):
```
canihavesomecoffee, line 190: Same

mod_api/routes/regression_tests.py
```diff
@@ -0,0 +1,349 @@
...
+    rows = query.offset(offset).limit(limit).all()
+    return paginated_response(
+        [_serialize_category(row) for row in rows], total, limit, offset)
+
+
+@mod_api.route('/categories', methods=['POST'])
+@require_roles(_WRITE_ROLES)
+@require_scope(Scope.RUNS_WRITE)
+@validate_body(CategoryCreateSchema)
+def create_category(validated_data=None):
```
canihavesomecoffee, line 283: Same

mod_api/routes/regression_tests.py
```diff
@@ -0,0 +1,349 @@
...
+    g.log.info(f'category {category.id} created via API by {g.api_user.id}')
+    return single_response(_serialize_category(category), http_status=201)
+
+
+@mod_api.route('/categories/<category_id>', methods=['PATCH'])
+@require_roles(_WRITE_ROLES)
+@require_scope(Scope.RUNS_WRITE)
+@validate_path_id('category_id')
+@validate_body(CategoryUpdateSchema)
+def update_category(category_id, validated_data=None):
```
canihavesomecoffee, line 309: Same

mod_api/routes/regression_tests.py
```diff
@@ -0,0 +1,349 @@
...
+
+    g.log.info(f'category {category.id} updated via API by {g.api_user.id}')
+    return single_response(_serialize_category(category))
+
+
+@mod_api.route('/categories/<category_id>', methods=['DELETE'])
+@require_roles(_WRITE_ROLES)
+@require_scope(Scope.RUNS_WRITE)
+@validate_path_id('category_id')
+def delete_category(category_id):
```
canihavesomecoffee, line 351: Same

canihavesomecoffee (changes requested):

canihavesomecoffee (approved):

PR #1172, opened 2026-08-11
Fix history endpoint timeout by paginating in SQL

### LINKED ISSUE #1161: GET /samples/{id}/history times out (504) on production data
# `GET /samples/{id}/history` times out (504) on production data

## Summary

The endpoint applies `limit`/`offset` in Python **after** loading and processing
every result for the sample, so the page size has no effect on how much work the
request does. On production this takes long enough to hit the gateway timeout.

## Reproduction

Against `sampleplatform.ccextractor.org`, with a `runs:read` token:

```bash
curl -H "Authorization: Bearer $TOKEN" \
  "https://sampleplatform.ccextractor.org/api/v1/samples/3/history?limit=5"
# HTTP 504 after 78.7s

curl -H "Authorization: Bearer $TOKEN" \
  "https://sampleplatform.ccextractor.org/api/v1/samples/7/history?limit=5"
# HTTP 504 after 61.4s
```

Both asked for 5 entries. Neither returned.

This does not reproduce on a small development database — with ~50 runs the
request is instant, which is presumably why it got through review.

## Cause

`mod_api/routes/samples.py`, `get_sample_history` (line 538):

```python
results = query.order_by(Test.id.desc()).all()          # L585 — every result, all runs
...
all_files = TestResultFile.query.options(               # L589 — files for every one of them
    joinedload(TestResultFile.regression_test_output)
    .joinedload(RegressionTestOutput.multiple_files)
).filter(TestResultFile.test_id.in_(test_ids)).all()
...
unique_tests = Test.query.filter(Test.id.in_(test_ids)).all()   # L602
_, timestamps_map = batch_get_run_data(unique_tests)            # L607
entries = _process_history_entries(...)

total = len(entries)                                    # L617
paged = entries[offset:offset + limit]                  # L618 — the slice happens here
```

For a sample exercised in most runs, `test_ids` is in the thousands, and the
`TestResultFile` query fans out further through the two nested `joinedload`s.
All of that runs regardless of `limit`.

The batching that is there (`batch_get_run_data`, `_preload_expected_outputs`)
is doing its job — it avoids N+1 — but it batches over the *whole* history
rather than over the page being returned.

## Suggested fix

Push the pagination into SQL, so the expensive follow-up queries only ever see
one page:

```python
total = query.count()
results = query.order_by(Test.id.desc()).offset(offset).limit(limit).all()
```

and drop the Python slice. Everything downstream (`test_ids`, `all_files`,
`unique_tests`, `timestamps_map`) then scales with the page, not with history.

One complication: `status` is currently filtered inside
`_process_history_entries`, after the entries are built, so `total` would no
longer be correct when `?status=` is supplied. Options, roughly in order of
preference:

1. Derive status in SQL so it can join the main query — best, but the biggest
   change, since status depends on result files and expected outputs.
2. Keep the load-everything path *only* when `status` is present, and use the
   paginated path otherwise. The unfiltered case is the common one and it is
   the one that currently 504s.
3. Cap how far back history is read (e.g. most recent N runs) and document it.

## Impact

Any API consumer that reads sample history. Concretely, it makes `sp
investigate --with-history` unusable against production: that command does one
history lookup per failing sample, so a run with 45 failures cannot complete.

## Environment

- `upstream/master` @ `fd0d5b3`
- Observed 2026-08-05 against the production deployment

devangdileep:
Fixes #1161

The `/samples/{id}/history` endpoint was loading all results into memory and then slicing with Python, which caused 504 timeouts on production when a sample had thousands of runs.

### What I changed
- When no `?status` filter is used, pagination now happens at the SQL level using `.offset()` and `.limit()` so only the needed page is fetched
- When `?status` is passed, the old behavior is kept since status is derived in Python
- Added an optional `?regression_test_id` filter to narrow history to a single test

### DIFF
--- a/mod_api/routes/samples.py
+++ b/mod_api/routes/samples.py
@@ -483,11 +483,41 @@ def _process_history_entries(
             'status': status,
             'platform': test.platform.value,
             'branch': test.branch,
-            'commit_sha': test.commit,
-            'tested_at': timestamps.get('completed_at') or timestamps.get('started_at'),
+            'tested_at': (
+                timestamps.get('completed_at')
+                or timestamps.get('started_at')
+            ),
             'failure_signature': failure_sig,
         })
     return entries
+
+
+def _build_history_entries(results, status_filter=None):
+    test_ids = list({r.test_id for r in results})
+    all_files = TestResultFile.query.options(
+        joinedload(TestResultFile.regression_test_output)
+        .joinedload(RegressionTestOutput.multiple_files)
+    ).filter(
+        TestResultFile.test_id.in_(test_ids)).all() if test_ids else []
+    files_by_result = defaultdict(list)
+    for f in all_files:
+        files_by_result[(f.test_id, f.regression_test_id)].append(f)
+
+    expected_by_rt = _preload_expected_outputs(results)
+
+    unique_tests = Test.query.filter(
+        Test.id.in_(test_ids)).all() if test_ids else []
+    test_map = {t.id: t for t in unique_tests}
+
+    _, timestamps_map = batch_get_run_data(unique_tests)
+
+    return _process_history_entries(
+        results,
+        files_by_result,
+        status_filter,
+        timestamps_map=timestamps_map,
+        test_map=test_map,
+        expected_by_rt=expected_by_rt)
 
 
 def _apply_history_filters(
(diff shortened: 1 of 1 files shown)


canihavesomecoffee:
Fixed via #1167. Your fix lacks unit tests and other things.

Issue #1173, opened 2026-08-13
Expected outputs roll forward from the previous run's actual output, so behavioural changes self-approve and PRs inherit phantom failures


cfsmp3:
## Summary

The file a regression test is compared against is not a fixed baseline. It is **whatever the previous run actually produced**, and only when that run's output *differed*. The result is a bistable comparison: the same code can pass or fail depending on which run a global pointer happens to name, behavioural changes get adopted as the new truth without anyone approving them, and PR branches inherit failures they did not cause.

## The code

`mod_ci/controllers.py:995-1050`:

```python
commit_name = 'fetch_commit_' + test.platform.value
commit_hash = GeneralData.query.filter(GeneralData.key == commit_name).first().value
last_commit = Test.query.filter(and_(Test.commit == commit_hash, Test.platform == test.platform)).first()
...
last_commit_files = db.query(last_files.c.got).filter(and_(
    last_files.c.regression_test_output_id == output_file.id,
    last_files.c.got.isnot(None)
)).first()
correct = etree.SubElement(file_node, 'correct')
if last_commit_files is None:
    correct.text = output_file.filename_correct                               # original baseline
else:
    correct.text = output_file.create_correct_filename(last_commit_files[0])  # previous run's ACTUAL output
```

`mod_test/models.py:386`:

```python
got = Column(Text(), nullable=True)  # If null/empty, it's equal to the expected version
```

`got` is populated only when a result mismatched. So the branch above selects the previous run's actual output when that run failed the test, and the original baseline when it passed.

## Observed behaviour

PR #2292 (ccextractor) removed a 40 ms contraction of teletext hide timestamps, changing every teletext cue end by one frame at 25 fps. Linux runs, in order:

| run | commit | compared against | pass | fail |
|---|---|---|---|---|
| 9428 | before #2292 | original baseline | 212 | 1 |
| 9434 | #2292 merge commit | original baseline (output now differs) | 167 | **46** |
| 9440 | after #2292 | 9434's `got` — the changed output | 212 | 1 |
| 9442 | after #2292 | 9434's `got` | 212 | 1 |
| 9476 | PR #2323 | pointer's run had `got = NULL` → original baseline | 167 | **46** |

[...]

PR #1179, opened 2026-08-16
[FIX] Incomplete runs must not report pass (#1177)

### LINKED ISSUE #1177: A run can report `pass` after executing almost none of its samples
### Summary

`_check_completed_run_status` iterates only the `TestResult` rows that exist. Samples that never ran produce no row, so they are invisible to it — a run reports `pass` whenever the few tests that did run happened to pass, regardless of how many never ran at all.

### Reproduction

No database or server needed; this calls the function on `master` directly:

```python
from mod_api.services import status

class FakeResult:
    def __init__(self, rt_id):
        self.regression_test_id, self.exit_code, self.expected_rc = rt_id, 0, 0

# One test produced a result. The other 236 never ran, so they have no rows.
verdict = status._check_completed_run_status(9360, {9360: [FakeResult(1)]}, {}, {})
print(verdict)   # -> 'pass'
```

### In production

| run | status | samples that ran |
|-----|--------|------------------|
| 9322 (PR 2279, linux) | `pass` | **1 of 474** |
| 9323 (PR 2279, windows) | `pass` | **1 of 474** |
| 9360 (PR 2292, linux) | `pass` | **1 of 237** |

PR #2279 is open right now and shows a passing run on both platforms.

The API also contradicts itself on the same run:

```
GET /runs/9360           ->  status: pass
GET /runs/9360/summary   ->  pass_count: 1, skipped_count: 236
```

Of 116 completed runs sampled, 8 had `skipped_count > 0`. The three above reported `pass`; the other five (9338, 9342, 9346, 9350, 9358 — between 70% and 87% executed) reported `fail`, because something that *did* run failed. So an incomplete run is reported as passing exactly when the surviving fragment happens to be green.

### Cause

```python
results = results_by_test.get(t_id, [])
if not results:
    return 'error'          # zero results is handled
for r in results:           # ... but this only sees rows that exist
    ...
return 'pass'
```

The zero-result case was considered explicitly; the partial-result case was not.

For context on how a run ends up like this: 9360's log shows `Starting with entry 1 of 1`, then two 504s, then log-upload retries — the runner stopped after the first category file. Its run config lists all 237 test ids, so the platform did intend to run them.

### Suggested fix

`mod_api/routes/runs.py:519` already derives the skipped count as `total_samples - len(results)`. The status derivation can use the same comparison and return `error` — or a new `incomplete` — when a run produced fewer results than it was configured to run.

### Impact

A pull request can show green having tested essentially nothing, which is how a bad change gets merged. Found while reviewing every open CCExtractor pull request through [sp_cli](https://github.com/CCExtractor/sp_cli), using read-only API access.

aryanorastar:
## Summary

- Fixes #1177: `_check_completed_run_status` only inspected existing `TestResult` rows, so a completed run that executed 1 of N samples could still report `pass` when that fragment was green.
- Completeness now compares reported regression-test IDs against the run’s expected sample set (customized selection, else all **active** regression tests — same rule as `/runs/{id}/summary` totals via shared `expected_regression_ids`).
- Missing expected samples → `error` (same class as zero results). Existing fail/pass checks still apply when the suite is complete.

## Test plan

- [x] `nose2 tests.api.test_services_status` (20 tests OK), including new `test_derive_run_status_completed_partial_results_is_error`
- [x] Existing pass/fail fixtures scoped with `CustomizedTest` so they remain meaningful under the completeness check
- [ ] CI green on this PR

### DIFF
--- a/mod_api/routes/runs.py
+++ b/mod_api/routes/runs.py
@@ -29,7 +29,8 @@ from mod_api.schemas.runs import (ProgressEventSchema, RunCreateRequestSchema,
                                   RunSchema, RunSummarySchema)
 from mod_api.services.error_service import derive_errors_for_run
 from mod_api.services.status import (batch_get_run_data, derive_run_status,
-                                     derive_sample_status)
+                                     derive_sample_status,
+                                     expected_regression_ids)
 from mod_api.utils import get_sort_column, paginated_response, single_response
 from mod_auth.models import Role
 from mod_customized.models import CustomizedTest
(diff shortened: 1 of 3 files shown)


aryanorastar:
Rebased onto current `master` (includes #1173 baseline comparison fix). Local `nose2 tests.api.test_services_status` still 20/20 green.

canihavesomecoffee (approved):

PR #1191, opened 2026-09-08
[FIX] Record the SHA CI actually built for PR runs

### LINKED ISSUE #1176: PR test runs record a commit that was never built
### Summary

For pull request runs, `Test.commit` stores the PR's head SHA, but the binary that actually runs is built from GitHub's merge ref — a merge of that head into master. The commit shown in the UI and returned by the API was never built.

### Evidence

Sampled 22 completed runs from the last 100, using read-only API access only:

|  | recorded commit == binary's own commit |
|---|---|
| master runs | **8 / 8** |
| PR runs | **2 / 14** |

The binary reports its own commit in the build log (`Git commit:` in the version block). For all 12 mismatching PR runs, that commit is a merge commit whose parents are exactly *(a master commit, the recorded head)*:

```
run 9490   recorded 6e1e22a3b8   binary e98f1a2f81
  e98f1a2f81  parents 128175ea7b + 6e1e22a3b8
  "Merge 6e1e22a3b84da8190629648690a49e7d9974d9ca into 128175ea7b..."

run 9486   recorded 73a7b90ac6   binary 0531cf3e8f
  0531cf3e8f  parents 128175ea7b + 73a7b90ac6

run 9474   recorded b9d6c45671   binary 5facb55537
  5facb55537  parents 6077cf5c61 + b9d6c45671
```

The two PR runs that *did* match (9492, 9493) fit the same explanation: their binary commit `aee9693292` has a single parent, `128175ea7b`, which was the master tip at the time. That branch was already even with master, so GitHub's merge ref resolved to the head itself and there was nothing to differ.

So: **the recorded and built commits diverge whenever GitHub produces a merge commit — that is, whenever the PR branch is not already sitting on the master tip.**

### Cause

`mod_ci/controllers.py:1994` takes the head SHA from the webhook payload:

```python
commit_hash = payload['pull_request']['head']['sha']
...
add_test_entry(g.db, commit_hash, TestType.pull_request, pr_nr=pr_nr)
```

That value is right as far as it goes — `schedule_tests` depends on it resolving through `repository.get_commit()` and on comparing it against `test_pr.head.sha`. The gap is that nothing records the commit that was actually built.

### Impact

1. **Reproduction gives you different code.** Checking out the commit shown on the run page and building it does not produce the binary that ran.
2. **Master's changes are present in the tested binary but not in the recorded commit.** A PR run's binary contains master commits that the PR's own base does not, so a comparison against a baseline at the recorded commit is not comparing what it appears to compare.

*Correction to an earlier version of this issue:* I originally cited "unrelated PRs reporting identical sets of new failures" as motivation here. That symptom is explained by #1173 — the expected file rolling forward from the previous run — not by this one. The commit mismatch above stands on its own evidence; I have removed the conflated claim.

### Suggested fix

Record the built commit alongside the head rather than instead of it — e.g. a `built_commit` column on `Test`, populated from the artifact's workflow run, or parsed from the `Git commit:` line the binary already prints. The UI can then show "PR head X, tested as merge Y", and API consumers can tell the two apart.

### Reproducing

```bash
sp run show 9490                              # recorded commit
sp run logs 9490 --all | grep 'Git commit'    # what the binary says
```

Found while reviewing every open CCExtractor pull request through [sp_cli](https://github.com/CCExtractor/sp_cli), without shell access to the server.

NaitikVerma6776:
Fixes #1176

## What was wrong

PR runs store the **PR head** SHA in `Test.commit`. GitHub Actions builds the **merge ref** (PR head merged into master), so the binary that actually ran is a different commit.

Run 9490 is the example Willem posted:

- `sp run show 9490` reports head `6e1e22a3b8`
- `sp run logs 9490 --all` has `Git commit: e98f1a2f81`

The UI/API therefore claimed we tested a commit that never went through the VM.

## What this PR does

- Keep `commit` as the PR head (the GitHub status / PR link still make sense).
- Add nullable `built_commit`.
- Fill it from `merge_commit_sha` when the PR webhook creates the run.
- Update it from the `Git commit:` line when the VM uploads logs (ground truth from the binary).
- Expose `built_commit_sha` on GET /api/v1/runs.
- Show `PR head …, tested as …` on the run page when the two SHAs differ.

Master/commit runs are unchanged (`built_commit` stays null unless logs say otherwise).

## Test plan

- [ ] Unit tests for parse / webhook / log upload / API field
- [ ] After deploy, compare `sp run show <pr-run>` `commit_sha` vs `built_commit_sha` against `sp run logs <id> --all | grep 'Git commit'`

### DIFF
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -392,7 +392,7 @@ def _merge_commit_from_pr(pr, payload_pr: Optional[dict] = None) -> Optional[str
     if payload_pr:
         candidates.append(payload_pr.get('merge_commit_sha'))
     for sha in candidates:
-        if is_valid_commit_hash(sha):
+        if isinstance(sha, str) and is_valid_commit_hash(sha):
             return sha.strip()
     return None
 
--- a/mod_ci/controllers.py
+++ b/mod_ci/controllers.py
@@ -304,6 +304,60 @@ def is_valid_commit_hash(commit: Optional[str]) -> bool:
         return False
 
 
+GIT_COMMIT_LOG_RE = re.compile(r'Git commit:\s*([0-9a-fA-F]{7,40})')
+
+
+def parse_git_commit_from_logs(text: Optional[str]) -> Optional[str]:
+    """
+    Extract the SHA the binary reports from uploaded VM logs.
+
+    GitHub Actions builds the PR merge ref, not the PR head. The tested
+    binary prints ``Git commit: <sha>``. That line is the ground truth for
+    what was actually compiled, as opposed to ``Test.commit`` which stores
+    the PR head SHA from the webhook.
+
+    :param text: Log file contents
+    :type text: Optional[str]
+    :return: A valid commit hash if found, otherwise None
+    :rtype: Optional[str]
+    """
+    if not text:
+        return None
+    match = GIT_COMMIT_LOG_RE.search(text)
+    if not match:
+        return None
+    sha = match.group(1)
+    return sha if is_valid_commit_hash(sha) else None
+
+
+def _merge_commit_from_pr(pr, payload_pr: Optional[dict] = None) -> Optional[str]:
+    """Return GitHub's test-merge SHA for a PR, if GitHub has computed one."""
+    candidates = []
+    if pr is not None:
+        candidates.append(getattr(pr, 'merge_commit_sha', None))
+    if payload_pr:
+        candidates.append(payload_pr.get('merge_commit_sha'))
+    for sha in candidates:
+        if is_valid_commit_hash(sha):
+            return sha.strip()
+    return None
+
+
+def _record_built_commit_from_log(log, test, log_path: str) -> None:
+    """Update Test.built_commit from a ``Git commit:`` line in uploaded logs."""
+    try:
+        with open(log_path, encoding='utf-8', errors='replace') as handle:
+            text = handle.read()
+    except OSError:
+        return
+    sha = parse_git_commit_from_logs(text)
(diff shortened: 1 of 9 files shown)


mod_ci/controllers.py
```diff
@@ -304,6 +304,60 @@ def is_valid_commit_hash(commit: Optional[str]) -> bool:
...
+        if is_valid_commit_hash(sha):
+            return sha.strip()
+    return None
+
+
+def _record_built_commit_from_log(log, test, log_path: str) -> None:
+    """Update Test.built_commit from a ``Git commit:`` line in uploaded logs."""
+    try:
+        with open(log_path, encoding='utf-8', errors='replace') as handle:
+            text = handle.read()
```
canihavesomecoffee: This is not going to work very well if it's a very large log file. Can you adjust so that it either reads in batches and returns as soon as it found something?

canihavesomecoffee (changes requested):

NaitikVerma6776:
@canihavesomecoffee ....Good catch ,  it now scans in 64KB batches and stops at the first Git commit: line, with overlap so a split line is still found. Please review the changes.

canihavesomecoffee (dismissed):

NaitikVerma6776:
hi @canihavesomecoffee CI is now passing on the latest commit.

The latest commit only addresses the CI/test issue around the mocked log path; the approved #1176 implementation remains unchanged.

Could you please re-review and approve the current HEAD when you get a chance? Thanks!

canihavesomecoffee (approved):

